PR #102, opened 2025-01-14
improve: adding MCP initial support for tools

ismaelfaro:
Initial support of MCP tools. Based in the [Bee TS support](https://github.com/i-am-bee/bee-agent-framework/blob/main/src/tools/mcp.ts)

### Dependencies

```bash
poetry --directory framework/bee-py install 
```

### Test it
```bash
pytest tests/bee_agent/tools/test_mcp_tool.py        
```


framework/bee-py/poetry.lock
```diff
@@ -1,4 +1,4 @@
-# This file is automatically @generated by Poetry 1.8.5 and should not be changed by hand.
```
planetf1, line 1: Do we need a policy on updating of the lock file.

Sometimes libraries don't check this in, but the benefit of checking in is consistency (especially for applications)
If we do want to go with checkins we may want to ensure consistent use of tool & python versions (as dependencies can resolve differently based on python version in use). One possibility (perhaps in future) is to do as part of a build process

In general though the updates are progressive and we're in early stages so no blocking concern.

framework/bee-py/bee_agent/tools/mcp_tools.py
```diff
@@ -0,0 +1,181 @@
+# SPDX-License-Identifier: Apache-2.0
+
```
planetf1, line 2: General point - how do we keep framework in sync (and ensure issues etc are fixed) in future. Will we address this as part of our overall repo refactoring? If we want to depend on select features in another repo we may want a mechanism to support this (subproject, review process, tagging etc).

framework/bee-py/bee_agent/tools/mcp_tools.py
```diff
@@ -0,0 +1,181 @@
...
+    ) -> Sequence[TextContent | ImageContent | EmbeddedResource]:
+        print(f"Multiply tool called with arguments: {arguments}")  # Debug
+        a, b = float(arguments["a"]), float(arguments["b"])
+        result = str(
+            int(a * b)
+        )  # Convert to int since we're testing with whole numbers
+        print(f"Multiply tool result: {result}")  # Debug
+        return [TextContent(type="text", text=result)]
+
+    async def run_tests():
```
psschwei, line 109: should we split this out into the test directory `framework/bee-py/tests` ?

ismaelfaro: yes. We can do it, I'm working is an agent to generate/extract the test and add them in the tests/ folder. 

@vabarbosa we can started to move all the tests there. Works for you?

framework/bee-py/poetry.lock
ismaelfaro: we can remove it for know to move it more agile, and define the policy on updating of the lock file.

framework/bee-py/bee_agent/tools/mcp_tools.py
vabarbosa: yes, we should

ismaelfaro: ok, let me start with this one.

ismaelfaro: done:

`pytest tests/bee_agent/tools/test_mcp_tool.py `

framework/bee-py/poetry.lock
ismaelfaro: I just remove it for now.

framework/bee-py/python_project.txt
```diff
@@ -0,0 +1,3108 @@
+
```
psschwei, line 1: How is this file used? (I don't think I've seen this paradigm before)

ismaelfaro: :), good catch, that is the concatenated file that I use to guide the agents to help me to make the tests. Let me remove it.

psschwei (approved):
tests pass!

Issue #42, opened 2024-11-25
platform: improve error logging


psschwei:
As a user, when one of my jobs fails I want to be able to diagnose and understand why it failed. To do this, I need better error messages and logs from the stack so that when there is an issue I know what the specific error is, which component caused it, and ideally can formulate an idea on how to resolve it.

Specifically, this is around errors in the workflow. (I could see this also surfacing errors raised by individual agents)

planetf1:
Similar comment to #43 - consider use of the [bee observability framework](https://github.com/i-am-bee/bee-agent-framework/blob/main/docs/instrumentation.md) which uses mflow

planetf1:
Currently the bee hive uses a simplistic execution engine. We could add observability support even into that implementation. This would help understand how we get the necessary dependency (which might be optional) in place, and start building the set of 'events' that we want to log (which is more than errors)

planetf1:
The open telemetry collector in the bee stack requires the additional `x-bee-authorization` header. The key for this is defined in the `docker-compose.yaml` ie `observe-auth-key`

planetf1:
- If we use open telemetry we can default/document a local configuration that only logs to files, and will have no dependence on the bee observability framework. The above/PR provides an approach we could adopt
- We could implement our own log/trace/observability class to add a level of abstraction ie independence from the current implementation

PR #103, opened 2025-01-14
improve: Add initial observability support

planetf1:
Initial support for Observability

 - uses open telemetry
 - requires the bee stack running with an active opentelemetry container (port 4318, localhost, hardcoded)
 
 Todo
 - enable/disable via config
 - endpoint via config
 - how do we hook into the same context as being used in the bee framework
 - currently using 'with trace' construct. Seems preferred in docs as helps to manage scope, though it does impact code layout. Annotations/explicit context are options too
 - need to figure out exactly what we want to record - spans, but also additional errors, events

This is currently work in progress & Presented for discussion to support issue #42 & related


maximilien (commented):
I like this but before going too far, can you please add a test @planetf1. At a minimum running the simplest agent and verifying that the basic Observability outputs are being done. Keep it simple :)

psschwei:
> I am assuming @psschwei or @akihikokuroda are reviewing this?

I generally don't look at draft PRs unless explicitly tagged (I usually wait until they move out of draft stage and into ready for review)

akihikokuroda:
I think that putting run_workflow.py is good for now to verify the configuration and outputs.  I assume that the actual configuration code eventually in the CLI or the service and the runtime code in agent, workflow, etc...

The multi agent flow user, the multi agent flow developer and the multi agent platform developer are the user for the observability.  They probably care about flow of among the agents and performance of the processing.   We have to design these users in mind.

psschwei:
we should implement this in the agent / workflow classes and functions (in agent.py and workflow.py) rather than in the standalone functions

planetf1:
This was a first pass to demonstrate and show my understanding of the tech used at the time.

Given the refactors, the comment (which I agree with) on integrating into workflow etc, plus knowledge about bee framework refactoring, will close this PR & consider appropriate approach moving forward

PR #105, opened 2025-01-14
pass base_url when inferencing

### LINKED ISSUE #104: Add `base_url` to `inference method in  `LLM` class
When using the `LLM` class, we don't pass the base_url

https://github.com/i-am-bee/bee-hive/blob/main/framework/bee-py/bee_agent/llms/llm.py#L86-L90

As a result, custom URLs are not supported (for example, local ollama when accessed through a container)

psschwei:
Fixes #104


vabarbosa (approved):
thank you

Issue #110, opened 2025-01-16
Implement sequential workflow as a function of Workflow class


psschwei:
xref https://github.com/i-am-bee/bee-hive/blob/c9f5b86829f4697511aaf9b240267873e4069785/bee-hive/bee_hive/run_workflow.py#L29-L41

psschwei:
One thought I had: the [sample workflow](https://github.com/i-am-bee/bee-hive/blob/c9f5b86829f4697511aaf9b240267873e4069785/examples/bee-hive/weather-checker/workflow.yaml) I used just listed agents in order and then looped through them. 

But when we get to parallel execution, we were thinking of a `dependsOn` (or something like that) to handle relationships between agents (see discussion in #23 ). So wondering if we want to use something similar here, in other words, creating a more explicit execution order rather than just a for loop? The upside would be sequential and parallel use the same execution under the hood, but the downside is it could be more complex to build now...

WDYT @akihikokuroda ?

akihikokuroda:
Dependency #106 and #107.

Issue #39, opened 2024-11-25
workflow language supports conditionals


psschwei:
Implement conditional workflow as function of Workflow class

akihikokuroda:
For now, Bee-hive doesn't see any output from agents.  Seeing the output from the agent may be necessary for this.  WDYT?

planetf1:
Have you had any thoughts about how to express conditions?

Some approaches have used preprocessing, or templating (ie helm) - but this gets away from yaml being the source

Currently we have an Agent and Workflow definition. The latter does have a strategy entry. I was wondering if we expect to add explicit task (or step, or both?) definitions within the workflow, so we end up with something more like github actions (or tekton?)

Just sketching around ?
```yaml
tasks:
  # A task is a 'unit' that probably executes with shared context?
  # is this needed?
  - name: getthenews
    # A step is an individual action that invokes an agent or similar
    steps:
      - name: current-affairs
        # explicitly refer to the agent we will invoke here
        execute: current-affairs
      - name: fact-check
        execute: fact-checker
        # introduce a condition (consider how we handle variables?)
        # this step is only used in the UK
        if: country == 'UK'
      - name: review
        # support matrix parallelism
        strategy:
          matrix:
            # we end up invoking with each value
            agent ['sme-reviewer','layperson-reviewer','investor-reviewer']
        execute: {{ matrix.agent }}
      - name: hot-or-not
        # How can we make use of variables within the executed agent? GHA uses shell script
        # where does 'number-of-hot-articles' come from? the execution must be able to modify it
        execute: hot-or-not
        until: number-of-hot-articles > 5
        # Thought - could we also use shell script? Adds variable expansion, loops etc?
        # but then that is programming ...
```

Is there a current proposal?

planetf1:
This also brings up the need to consider
 - variable management
 - parsing of expressions
 - input/outputs
 - error conditions
 
 Since there are some examples out there of yaml definitions (with open licenses) which have tried this, perhaps we could take inspiration from these as the above seem common considerations?

akihikokuroda:
How about making the agent decide the next step?  

This is in "hot-or-not" agent.

> instructions: The user will give you a temperature in Fahrenheit and a location. Use the OpenMateo weather tool to find the average monthly temperature for the location. Answer if the temperature provided by the user is hotter or colder than the average found by the tool.

it returns `hotter` or `colder`. This answer can be used as the name of the next step.

psschwei:
[...]

PR #120, opened 2025-01-16
poc1 skelton

akihikokuroda:
I've created a base of POC1 from the current main branch files.  This PR can be merged or just used for reference.

### DIFF
--- a/bee-hive/bee_hive/__init__.py
+++ b/bee-hive/bee_hive/__init__.py
@@ -1,0 +1,4 @@
+__all__ = [
+    "agent",
+    "workflow",
+]
--- a/bee-hive/bee_hive/agent.py
+++ b/bee-hive/bee_hive/agent.py
@@ -1,0 +1,112 @@
+#! /usr/bin/env python3
+# SPDX-License-Identifier: Apache-2.0
+
+import json
+import os
+import sys
+
+import dotenv
+from openai import OpenAI
+import yaml
+
+dotenv.load_dotenv()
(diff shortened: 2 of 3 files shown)


akihikokuroda:
I wonder if we may create a "poc1" branch in bee-hive so that we can put and share incomplete code in the repository while we are working for the POC1.

psschwei:
> create a "poc1" branch so that we can put and share incomplete code

I'd lean more towards checking in complete code, but breaking it into smaller PRs so each commit can be complete. 

so for example if there are going to be 5 functions in the workflow class, I'm totally on board with 6 different PRs (one to create the class and one each for the different functions)

heck, just adding the directory structure you posted in slack (with no additional added code) would be fine as a PR.

(added benefit of small PRs - easier to review :smile: )

maximilien (commented):
You probably need to sync up with @psschwei Aki. Maybe you both can split the work. 

Also we should drive this from the workflow YAML so perhaps that’s the first step.

Can you share the YAML? If you don’t have that then let’s focus on creating that too.

The point is the YAML would be the use case and be where the agents and workflows are defined.

akihikokuroda:
I can not work for #110 and  #39 with the current code in `main` branch.  The `Workflow` class and `Agent` class definition must be merged first.

akihikokuroda:
The assumption is that the yaml syntax is same as the POC0 for now in this PR.

akihikokuroda:
This PR is for unblock the work for #39 and #110, etc..

psschwei (approved):

PR #123, opened 2025-01-17
Update sequence workflow

### LINKED ISSUE #110: Implement sequential workflow as a function of Workflow class

akihikokuroda:
fix #110 

example of the workflow class and run method is 
```
def parse_yaml(file_path):
    return yaml_data

agents_yaml = parse_yaml("agent.yaml")
workflow_yaml = parse_yaml("workflow.yaml")
try:
    workflow = Workflow(agents_yaml, workflow_yaml[0])
except Exception as excep:
    raise RuntimeError("Unable to create agents") from excep
workflow.run()
```

The yaml schema is the same as poc 0.

### DIFF
--- a/bee-hive/bee_hive/workflow.py
+++ b/bee-hive/bee_hive/workflow.py
@@ -29,18 +29,18 @@ class Workflow:
         """Execute workflow."""
 
         if (self.workflow["spec"]["strategy"]["type"]  == "sequence"):
-            prompt = self.workflow["spec"]["template"]["prompt"]
-            for agent in self.agents:
-                if (
-                    self.workflow["spec"]["strategy"]["output"]
-                    and self.workflow["spec"]["strategy"]["output"] == "verbose"
-                ):
-                    prompt = agent.run_streaming(prompt)
-                else:
-                    prompt = agent.run(prompt)
-            return prompt
+            return self._sequence()
         else:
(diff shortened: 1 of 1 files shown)


bee-hive/bee_hive/workflow.py
```diff
@@ -9,37 +9,38 @@
...
-           ):
-            run_workflow = run_streaming_agent
+            self.agents.append(Agent(agent_def))
+        self.workflow = workflow
+
+
+    def run(self):
+        """Execute workflow."""
+
+        if (self.workflow["spec"]["strategy"]["type"]  == "sequence"):
```
psschwei, line 31: would it make sense to pull this into a hidden helper function? something like:

```python
class Workflow:
    def run(self):
        if strategy == sequence:
            return _run_sequence(workflow)

    def _run_sequence(workflow):
        ...
```

akihikokuroda: I agree. I was going to do when other types are added.  I do it now.

psschwei (approved):
thanks Aki!

Issue #48, opened 2024-11-25
agent: schema


psschwei:
As a user, I want to have a formal spec declared for agents. This spec should include sections for:

- [ ] tools
- [ ] logic
- [ ] inputs / outputs

this spec must adhere to a defined schema

psschwei:
> What is the "logic"? Is it some custom code?

More like the system prompts - how should the agent "think" about the job its supposed to do

> Is `input` the `output` of the previous step? Is there anything else? 
> Is `output` the text part of the LLM output?

IIRC the thinking here was around the structure/template of the inputs and outputs, and also down the line potentially some type of allow/deny list for who was allowed to communicate with an agent.

akihikokuroda:
There is an `instructions`.  Can it be the `logic`?
The `input` and `output` can be `string` for now?

psschwei:
Yeah, I think both of those are fine

Issue #18, opened 2024-11-25
workflow language supports parallel execution of agents


psschwei:
Part of #5 

As a user, I can invoke agents in parallel when building a workflow. 

"Agents" here can mean both single agents or a workflow of multiple agents.

akihikokuroda:
@psschwei May I take this as a low priority item for now?

psschwei:
👍

akihikokuroda:
Making steps (my assumption of the parallelized unit) in parallel is easy but input, output and dependencies must be designed.

- The step takes one input and generates an output and the next step name.  They are saved until they are used by the step depending on this step and match as the next step as their input.
- Each step may have one or more dependencies.  When the workflow is started, the steps without dependencies are started as the prompt of the workflow as the input.
- Steps with dependencies are executed when all the dependent steps are executed and the next step matches the step.  The number of input of the step with dependencies is the same as the number of dependencies.
- At this moment, I assume that all of dependencies must be satisfied.   One of dependencies may required.

akihikokuroda:
For the implementation side, to support **parallel**, the workflow probably needs to be started by `asyncio.run` and most of current methods in `agent.py` and `workflow.py` become `async` and called with `await`.

PR #124, opened 2025-01-20
add condition

### LINKED ISSUE #39: workflow language supports conditionals

akihikokuroda:
fix #39 

This is a proposed implementation of the **condition.**. 

There is a new object `step` that has `name`, `condition`, `agent` and `dependencies`. 
Here is an example yaml file.  The condition is a python code that generate `output` object that has `prompt` and `next` items.  The `prompt` is a text output for the next step.  The `next` is the name of the next step.
Step in the `spec.template.start` is executed first.  The next step is the `next` in the `output` of the step execution.  The 'end' in the `next` finish the execution.  **The python execution for the condition evaluation is not safe.  It may need to be replaced some other way**

```
kind: Workflow
metadata:
  name: beehive-deployment
  labels:
    app: mas-example
spec:
  strategy:
    type: condition
    output: nonal
  template:
    metadata:
      labels:
        app: mas-example
    agents:
      - name: current-affairs
      - name: hot-or-not
    prompt: New York
    start: current-affairs
    steps:
      - name: current-affairs
        agent: current-affairs
        condition: |
            print("in condition")
            print(prompt)
            output = {"prompt": prompt, "next": "hot-or-not"}	    
        dependencies:
      - name: hot-or-not
        agent: hot-or-not
        condition: |
            print("in condition")
            print(prompt)
            output = {"prompt": prompt, "next": "end"}	    
        dependencies:
```

Here is the example program that executes the example workflow.

```
import json
import os
import sys

import dotenv
from openai import OpenAI
import yaml
from workflow import Workflow

dotenv.load_dotenv()


def parse_yaml(file_path):
    return yaml_data

if __name__ == "__main__":
    agents_yaml = parse_yaml("agent.yaml")
    workflow_yaml = parse_yaml("workflow.yaml")
    try:
        workflow = Workflow(agents_yaml, workflow_yaml[0])
    except Exception as excep:
        raise RuntimeError("Unable to create agents") from excep
    workflow.run()
```

### DIFF
--- a/bee-hive/bee_hive/agent.py
+++ b/bee-hive/bee_hive/agent.py
@@ -31,11 +31,13 @@ class Agent:
         self.agent_output = agent["spec"].get("output")
         self.agent_tools = []
 
-        for tool in agent["spec"]["tools"]:
-            if tool == "code_interpreter":
-                self.agent_tools.append({"type": tool})
-            else:
-                print(f"Enable the {tool} tool in the Bee UI")
+        tools = agent["spec"].get("tools")
+        if tools:
+            for tool in tools:
+                if tool == "code_interpreter":
+                    self.agent_tools.append({"type": tool})
+                else:
+                    print(f"Enable the {tool} tool in the Bee UI")
 
         instructions = f"{agent_instr} Input is expected in format: {self.agent_input}" if self.agent_input else self.agent_instr
         instructions = f"{instructions} Output must be in format: {self.agent_output}" if self.agent_output else instructions
--- a/bee-hive/bee_hive/step.py
+++ b/bee-hive/bee_hive/step.py
@@ -1,0 +1,34 @@
+#! /usr/bin/env python3
+# SPDX-License-Identifier: Apache-2.0
+
+import json
+import os
+import sys
+
+import dotenv
+from openai import OpenAI
+import yaml
+
+dotenv.load_dotenv()
+from agent import Agent
+
+class Step:
+    def __init__(self, step):
+        self.step_name = step["name"]
+        self.step_agent = step.get("agent")
+        self.step_condition = step.get("condition")
+        self.step_dependencies = step.get("dependencies")
+
+    def run(self, prompt):
+        output = {}
+        if self.step_agent:
+            prompt = self.step_agent.run(prompt)
+            output["prompt"] = prompt
+        if self.step_condition:
+            input = {"prompt": prompt}
+            exec(self.step_condition, {}, input)
+            output = locals()["input"].get("output")
+            print(output)
+        if self.step_dependencies:
+            output["dependencies"] = self.step_dependencies
+        return output
--- a/bee-hive/bee_hive/workflow.py
+++ b/bee-hive/bee_hive/workflow.py
@@ -10,9 +10,11 @@ import dotenv
 
 dotenv.load_dotenv()
 from agent import Agent
+from step import Step
 
 class Workflow:
-    agents = []
+    agents = {}
+    steps = {}
     workflow = {}
     def __init__(self, agent_defs, workflow):
         """Execute sequential workflow.
@@ -21,7 +23,7 @@ class Workflow:
             workflow: workflow definition
         """
         for agent_def in agent_defs:
-            self.agents.append(Agent(agent_def))
+            self.agents[agent_def["metadata"]["name"]] = Agent(agent_def)
         self.workflow = workflow
 
 
@@ -30,12 +32,14 @@ class Workflow:
 
         if (self.workflow["spec"]["strategy"]["type"]  == "sequence"):
             return self._sequence()
+        elif (self.workflow["spec"]["strategy"]["type"]  == "condition"):
+            return self._condition()
         else:
             print("not supported yet")   
 
     def _sequence(self):
         prompt = self.workflow["spec"]["template"]["prompt"]
-        for agent in self.agents:
+        for agent in self.agents.values():
             if (
                 self.workflow["spec"]["strategy"]["output"]
                 and self.workflow["spec"]["strategy"]["output"] == "verbose"
@@ -43,4 +47,18 @@ class Workflow:
                 prompt = agent.run_streaming(prompt)
             else:
                 prompt = agent.run(prompt)
+        return prompt
+
+    def _condition(self):
+        prompt = self.workflow["spec"]["template"]["prompt"]
+        steps = self.workflow["spec"]["template"]["steps"]
+        for step in steps:
+            if step["agent"]:
+                step["agent"] = self.agents.get(step["agent"])
+            self.steps[step["name"]] = Step(step)
+        current_step = self.workflow["spec"]["template"]["start"]
+        while current_step != "end":
+            response = self.steps[current_step].run(prompt)
+            prompt = response["prompt"]
+            current_step = response["next"]
         return prompt


akihikokuroda:
Here is an example that demonstrates the condition.

joke_agents.yaml
```
apiVersion: beehive/v1alpha1
kind: Agent
metadata:
  name: expert
  labels:
    app: mas-example
spec:
  model: "llama3.1:latest"
  description: expert
  instructions: Say a joke.

---
apiVersion: beehive/v1alpha1
kind: Agent
metadata:
  name: colleague
  labels:
    app: mas-example
spec:
  model: "llama3.1:latest"
  description: colleague
  instructions: your colleague likes joks.  He says a joke. Put score for the joke between 1-10, 10 is the funniest.  If the score is 10, say the score and the joke otherwise say "next". 
```
joke_workflow.yaml
```
apiVersion: beehive/v1alpha1
kind: Workflow
metadata:
  name: beehive-deployment
  labels:
    app: mas-example
spec:
  strategy:
    type: condition
    output: nonal
  template:
    metadata:
      labels:
        app: mas-example
    agents:
      - name: expert
      - name: colleague
    prompt: Tell me a joke about IBM
    start: expert
    steps:
      - name: expert
        agent: expert
        condition: |
            print("in condition")
            print(prompt)
            output = {"prompt": prompt, "next": "colleague"}	    
        dependencies:
      - name: colleague
        agent: colleague
        condition: |
            print("in condition")
            print(prompt)
            if (prompt.find("next") == -1):
                output = {"prompt": prompt, "next": "end"}
            else:
                output = {"prompt": "Can you give me a funnier joke about IBM?", "next": "expert"}                
        dependencies:%                                                    
```
test_jokeworkflow.py
```

import json
import os
import sys

import dotenv
from openai import OpenAI
import yaml
from workflow import Workflow

dotenv.load_dotenv()

def parse_yaml(file_path):
    return yaml_data

if __name__ == "__main__":
    agents_yaml = parse_yaml("joke_agents.yaml")
    workflow_yaml = parse_yaml("joke_workflow.yaml")
    try:
        workflow = Workflow(agents_yaml, workflow_yaml[0])
    except Exception as excep:
        raise RuntimeError("Unable to create agents") from excep
    workflow.run()

```

maximilien:
Nice. This is the conditional framework I was thinking of, see if you can reconcile your approach with this. @psschwei also suggested an additional which I will list after.

```yaml

apiVersion: beehive/v1
kind: Workflow
metadata:
  name: beehive-deployment
  labels:
    app: example2
spec:
  strategy:
    type: sequence
  template:
    metadata:
      labels:
        app: mas
        use-case: cbom
    agents:
        - finder
        - fixer
        - writer
    steps:
        exception:
            - step4
        step1:
            finder:
                code: |
                # code here
        step2:
            condition:
                if: |
                # expression
                then:
                    step1: |
                    # additional code
                else: |
                    # additional code
        step3:
            writer:
                code: |
                # code here
        step4:
            code: |
            # exception code
```

maximilien (commented):
See my comment. If we can resolve and agree then I will finish the review of this.

Thanks and sorry for asking schema question here. I have a PR for schema issue #48 that I am working on but don’t want to wait on that for agreement since we can slowly build the agreement.

akihikokuroda:
The differences are the definition of `step`.  Is there schema of the step? What are the input and output of the step?  
In this PR, I assume that the input is the prompt string and the output is the result string and the name of the next step (or `end`).  The step does: 
1. run an agent with the prompt (previous step result string)
2. check the result string from the agent and compose a result string and decide next step.

Thanks!

developer-gliu:
There is a suggestion I want to point out based upon my own experience with running the current bee framework. 

When we have pretty complex tasks, it can take a while to actually return a result. For example, in the CBOM demo, one of the agents took over 2 minutes. Would it make sense to add a `timeout`field to handle execution time limits, and if surpassed, go to the exception? (which I think @maximilien had defined in his schema)

planetf1:
* @akihikokuroda For expressions, we could consider using 'eval' vs exec. A custom parser may be more secure, but more work/less flexible
* @maximilien where you have *code* in your proposal - is this python code? (similarly to how in github actions/tekton we end up frequently with shell script). Where does the agent get invoked? anywhere from the code, or just once during the step?
* To throw another suggestion in - see also [this post](https://github.com/i-am-bee/bee-hive/issues/39#issuecomment-2587165786) I added on one possible way - more gh action like
* Whenever we have a *sequence* that needs preserving (like steps defined in order), we should use a hyphen, not just mappings ( see [spec](https://yaml.org/spec/1.2.2/#741-flow-sequences) or a [more readable version](https://www.redhat.com/en/blog/yaml-explained-for-rosa-users) - this would specifically apply to *steps*

psschwei:
> What are the input and output of the step?

That's a good question, I don't think we've defined that in the schema yet (?)

developer-gliu:
> > Is there schema of the step?
> 
> I think this is the key thing we need to work out.
> 
> I'd propose a hybrid between both examples:
> 
> ```yaml
> steps:
>   - name: step1
>     condition:
>       - if: some condition
>         then: do something
>         else: do something else
>   - name: step2
>     condition:
>       - case: some condition
>         do: do something
>       - case: some other condition
>         do: do something else
>       - default:
>           do: do default something
> ```
> 
> key points:
> 
> * steps is a list of steps
> * step name needs to be configurable (not hardcoded as step1, step2, ... )
> * conditions should be templated rather than pure code blocks (so `if: /*code*/ then: /*code*/` rather than `condition: /*if/then code */`)

I'm having some trouble understanding how we would define input/output in a schema like this, for example, lets assume we use this hybrid approach and have a schema like this:

```
steps:
  - name: step1
    input: 
      - input_variable_1
      - input_variable_2
    output: 
      - output_variable_1
    condition:
      - if: some condition
        then: do something
        else: do something else
  - name: step2
    input: 
      - output_variable_1  
    output: 
      - output_variable_2
    condition:
      - case: some condition
        do: do something
      - case: some other condition
        do: do something else
      - default:
          do: do default something
```

maybe for different types of queries we would have a different number of inputs (as well as different possible types). How would we handle this?

akihikokuroda:
>   - name: step1
>     condition:
>       - if: some condition
>         then: do something
>         else: do something else
>   - name: step2
>     condition:
>       - case: some condition
>         do: do something
>       - case: some other condition
>         do: do something else
>       - default:
>           do: do default something

I want to simplify the step.  It can have one of:
1. one condition that decides next step name
2. one execution (agent interaction)
3. one execution and one condition

The condition takes one input (previous step output or the output of the execution in the same step).

The input is the output of the previous step (so one input until we do parallelism)
The output is the output of the execution and the next step name (default next step must be defined for the step that doesn't have the condition)

> conditions should be templated rather than pure code blocks (so if: /*code*/ then: /*code*/ rather than condition: /*if/then code */)
This is fine.  It can put the logic in one function so changing the template doesn't affect the main logic if the step is simplified.

WDYT

akihikokuroda:
When #18 is implemented the 
> one execution (agent interaction)

probably is changed 
> one execution (agent interaction or workflow invocation)

and the parallelized unit is the `step`.

psschwei:
> I want to simplify the step. It can have one of:
>   1. one condition that decides next step name
>   2. one execution (agent interaction)
>   3. one execution and one condition

works for me

psschwei:
@akihikokuroda does this need a review or are you still working on it?

akihikokuroda:
It's ready for review.  The implementation probably need to be refactored when we agree on the condition schema, though.
Thaks!

psschwei (approved):
:shipit:

maximilien:
OK folks. I think we converged to this:

```yaml
steps:
  - name: step1
    condition:
      - if: some condition
        then: do something
        else: do something else
  - name: step2
    condition:
      - case: some condition
        do: do something
      - case: some other condition
        do: do something else
      - default:
          do: do default something
```

And PR # adds the necessary schema for this change. Please review

PR #125, opened 2025-01-20
(DRAFT) feat(agents): Add support to workflow engine for third party agents

planetf1:
**WORK IN PROGRESS**

This PR adds support for alternate agent runners
 - Crew.AI

 in addition to
 - Bee
 
 Caveats
  - Actual agents implemented in i-am-bee/bee-hive#121 (and will need some refactoring to work with this)
  - still need to load/access/place 3rd party agent code in `$PYTHONPATH` for execution
  - Uses agent name to determine technology. Can consider how to define tech in a later PR, perhaps a type in the yaml. But maybe we need more metadata or some packaging for agents. See i-am-bee/bee-hive#130
  - existing bee code not refactored. See i-am-bee/bee-hive#128
  - no test cases. See i-am-bee/bee-hive#127 for framework, i-am-bee/bee-hive-demos#5 for simple example
  - does not address work to define input/output schema 
  - does not split the class definitions to separate files - but if this approach looks ok it would be a logical step. See i-am-bee/bee-hive#128
  - does not implement a langgraph runner. See i-am-bee/bee-hive#129
  
 Design
  - Add abstract base class for an agent runner
  - Add classes based on above for each technology
  - Adds only init()/run()/stream
  - connect up existing bee implementation
  - implement a crewai runner

Testing
 - bee agent (weather) runs as before 
 - crewai agent runs from sample workflow (will be added in i-am-bee/bee-hive-demos#5)

### DIFF
--- a/bee-hive/bee_hive/run_workflow.py
+++ b/bee-hive/bee_hive/run_workflow.py
@@ -71,6 +71,9 @@ class CrewAIAgentRunner(AgentRunner):
         # Implement CrewAI agent execution logic here
         # crew = Crew(agent_name, ...)
         # crew.kickoff() << this is how we kick it off in the sample code. Note could be additional tasks
+        
+        # TODO: We get called with prompt='New York' or similar. test agent needs { "location": "New York" }
+        #prompt2  = { "location": prompt }
         try: 
             # TODO: need to lookup method to call - base don agent name?
             self.instance.activity_crew().kickoff(prompt)
--- a/bee-hive/bee_hive/run_workflow.py
+++ b/bee-hive/bee_hive/run_workflow.py
@@ -274,10 +274,15 @@ def sequential_workflow(workflow):
 
     for agent in agents:
         try:
+
             agent_runner = AgentRunnerFactory.create_agent_runner(agent["name"])
-            # TODO: reinstate streaming support
-            #prompt=agent_runner.stream(prompt) if( workflow_yaml["spec"]["strategy"]["streaming"] and workflow_yaml["spec"]["strategy"]["output"] == "verbose" ) else agent_runner.run(prompt)
-            agent_runner.run(prompt)
+
+            if ( workflow_yaml["spec"]["strategy"]["output"]
+                 and 
+                 workflow_yaml["spec"]["strategy"]["output"] == "verbose" ):
+                prompt = agent_runner.stream(prompt)
+            else:
+                prompt = agent_runner.run(prompt)                
         except ValueError as e:
             print(e)
             raise(e)
--- a/bee-hive/bee_hive/run_workflow.py
+++ b/bee-hive/bee_hive/run_workflow.py
@@ -50,17 +50,18 @@ class CrewAIAgentRunner(AgentRunner):
     def __init__(self, agent_name):
         super().__init__(agent_name)
         # load python module based on name
-        # TODO: improve flex. for now agent_name should be the module + classname (which must begin with crewai_)
-        # For example:
-        #   crewai_things_to_do_cold_weather.ColdWeatherCrew
+        # TODO: improve configuration - current requirement is to have the agent name in the format:
+        #   <directory>.<filename>.<class>.<method> where 'crewai_' must appear in the name ie
+        #   test.crewai_test.ColdWeatherCrew.activity_crew
         try:
-            module_name, class_name = agent_name.rsplit(".", 1)
+            partial_agent_name, method_name = agent_name.rsplit(".", 1)
+            module_name, class_name = partial_agent_name.rsplit(".", 1)
             my_module = importlib.import_module(module_name)
             # Get the class object
             self.crew_agent_class = getattr(my_module, class_name)
-
             # Instantiate the class
-            self.instance = self.crew_agent_class()
+            self.instance = self.crew_agent_class()            
+            self.method_name = method_name
         except Exception as e:
             print(f"Failed to load agent {agent_name}: {e}")
             raise(e)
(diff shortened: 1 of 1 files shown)


planetf1:
method name is now also picked up from agent_name (short term implementation is based on naming convention)

bee-hive/bee_hive/run_workflow.py
```diff
@@ -25,19 +31,129 @@ def load_agent(agent):
...
     return agent_store[agent]
 
-
+# --- Abstract Base Class for Agent Runners ---
+class AgentRunner(ABC):
+    def __init__(self, agent_name):
+        self.agent_name = agent_name
+
+    @abstractmethod
+    def run(self, prompt):
```
maximilien, line 40: Can we enforce a return value for the `run(..) -> dict`?

bee-hive/bee_hive/run_workflow.py
```diff
@@ -25,19 +31,129 @@ def load_agent(agent):
...
+        """Runs the agent with the given prompt."""
+        pass
+    
+    @abstractmethod
+    def stream(self, prompt):
+        """Runs the agent in streaming mode with the given prompt."""
```
maximilien, line 46: Add some comments to at least the interface methods `run` and `stream`

bee-hive/bee_hive/run_workflow.py
```diff
@@ -25,19 +31,129 @@ def load_agent(agent):
         agent_store = json.load(f)
     return agent_store[agent]
 
-
+# --- Abstract Base Class for Agent Runners ---
+class AgentRunner(ABC):
```
maximilien, line 35: We need a generic way to deal with exceptions. Right now it seems like you are just passing the exception to the caller. 

That’s one approach. Another approach to think about is an exception handler. So something like:

```python
class AbstractAgent(ABC):
  #…
  def exception_handler(self, handler) -> error:
     pass
  #…
```
We can then let the `handler` object be defined and passed by the caller. This will allow exceptions to be handled by caller of the runner. For instance if there was a policy to retry.

Also if the handler is not defined then by default the exception is propagated up as you are doing now.

maximilien (changes requested):
Please also add a simple test :)

See other feedback., But this is definitely a step in a cool direction.

Thanks for doing this Nigel 🍻

bee-hive/bee_hive/run_workflow.py
planetf1: yes, still need to do this.

bee-hive/bee_hive/run_workflow.py
planetf1: We need comments in the whole module. Actually I think we need to split the file soon which will improve clarity & reduce edit conflicts
I plan to add for the additional code I added

bee-hive/bee_hive/run_workflow.py
planetf1: Agreed - inclined to promote this to an issue in it's own right so we can have a dedicated discussion on how we manage exceptions

planetf1:
> Please also add a simple test :)

We need to discuss test strategy. Will open an issue.
What does unit testing mean? Does it include the LLM? How we we execute that locally? ci?

bee-hive/bee_hive/run_workflow.py
planetf1: Added
- #131 : simple workflow for ad-hoc testing
- #127 : automated testing (local/pr/ci)

psschwei (changes requested):
We should make these changes in either `workflow.py` or `agent.py` - those are the files we'll use moving forward.

I used `run_workflow.py` and `create_agents.py` to hack together the initial demo, but they're not robust -- once we have all the functionality in the other files and the demos using them we'll get rid of them

sorry, don't think I had said that generally before

planetf1:
@psschwei I don't see a `workflow.py` or `agent.py` yet. Is this work in progress in another PR? I wasn't aware of the change, or the plan for 'other files' ? I do agree they do need to be more robust & split up, but I may need to wait until that new code is added before I can add in this change. How far are we away from that, since I'll be blocked on getting this in.

bee-hive/bee_hive/run_workflow.py
planetf1: Will track in #134 - so we can get the basics refactored merged asap since the base code layout is changing.

planetf1:
Looking at the refactored agent/workflow code, it's following a similar pattern to where I was going, except that:
 - Agent is a bee specific implementation, which approximately corresponds to the BeeAgentRunner in this PR
 - use of inheritance (base abstract class)
 
 Would it therefore make sense for us (in principle) to
 - still have the abstract base class (AgentRunner) - perhaps simply called Agent, or AgentBase
  - move the existing Agent implementation in BeeAgent to BeeAgentRunner
  - Add in CrewAIAgent / LangchainAgent
  
  I'm not quite sure yet how to handle the multiple implementations in workflow.yaml. It would be easier if we knew the type of the agent from the workflow, but this has a schema impact, so can probably use the existing assumptions around name.
  
  It does also mean needing to get clarity over what attributes are kept in the base class vs the Bee implementation.
  I can split these into multiple PRs to replace the existing one.
  
  An alternative is to just move the runners (and the conditional check) straight into Agent (similar to how they are in this pr in run_workflow.py), but this seems to at best to be less clean, and will likely lead to more conditional behaviour within the Agent code ?
  
  cc: @psschwei @akihikokuroda I'll look at scheduling a meeting to discuss

Issue #27, opened 2024-11-25
Schema: define basic schema for workflow


psschwei:
As a user, I want a defined schema to write my beehive workflow in. 

This is related with #47 and #48

maximilien:
Did you upload the example workflow that you presented in demo @psschwei?

Also similar question @developer-gliu?

I want to ensure schema matches. ideally we should have these in a `demo` or `use-case` directory. LMK

developer-gliu:
Yup, I have it in a branch currently. I don't exactly have a demo folder, but I just followed pauls current structure. 
The hierarchy is the examples -> pauls example folder / my example folder -> code
<img width="358" alt="Screenshot 2025-01-06 at 9 24 57 PM" src="https://github.com/user-attachments/assets/d34d5a56-6a13-4c41-a2f2-e5bb1bd7d9ac" />

psschwei:
> Did you upload the example workflow that you presented in demo @psschwei?
Yes, it's in a top-level `examples` directory, though that may change as there's some discussion about having separate examples directories for each subproject rather than a single one for all

maximilien:
OK folks, so for the schema I did some research and the best approach is to use JSON schema. Since YAML can be converted to JSON then the same JSON schema can be used to validate the YAML. This is my first attempt:

Proposed schema for the current agreed upon workflow (does not include conditional since we need to agree on that. See my comment in PR #124 

```json
{
  "$schema": "https://json-schema.org/draft/2020-12/schema",
  "$id": "https://github.com/i-am-bee/bee-hive/workflow.schema.json",
  "title": "Bee Hive",
  "description": "A schema for defining Bee Hive workflows in YAML or JSON",
  "type": "object",
  "properties": {
    "apiVersion": "string",
    "kind": "string",
    "metadata": {
      "name": "string",
      "labels": "object"
    },
    "spec": {
      "strategy": {
        "type": "sequence"
      },
      "template": {
        "metadata": {
          "name": "string",
          "labels": "object"
        }
      }
    },
    "agents": [],
    "exception": "string",
    "steps": []
  }
}

```

A valid JSON and YAML for this schema follows:

```json
{
  "apiVersion": "beehive/v1",
  "kind": "Workflow",
  "metadata": {
    "name": "beehive-deployment",
    "labels": {
      "app": "example2"
    }
  },
  "spec":{
    "strategy": {
      "type": "sequence"
    },
    "template": {
      "metadata": {
[...]

Issue #132, opened 2025-01-22
CLI "validate" function for workflows


maximilien:
Be able to validate workflows remotely via the CLI using a YAML file

PR #133, opened 2025-01-22
first version of the JSON schema. Fixes issue #27

maximilien:
Simple JSON workflow that adds the basic workflow features in RC0. One addition is the definition of the "exception" step.

There are two test files in `test/schmas/simple.<yml|json>`

Plan is to have tests for issue #132 that will use this file.

You can manually verify the sample tests files using: https://jsonschema.dev

JSON to YAML conversion can be done easily with `json` and `yaml` libraries or using: https://www.bairesdev.com/tools/json2yaml/ or similar tool

CC: @akihikokuroda @psschwei please do a quick review and merge so I can push the CLI code. Thx


test/schema/simple.yaml
```diff
@@ -0,0 +1,34 @@
...
+  - step3
+  - step4
+  step1:
+    agent: agent1
+  step2:
+    agent: agent2
+  step3:
+    agent: agent1
+  step4:
+    agent: agent3
```
psschwei, line 34: Instead of hardcoding the step names, could we do something like this instead?

```yaml
steps:
- name: step1
  agent: agent1
- name: step2
  agent: agent2
```

workflow_schema.json
```diff
@@ -0,0 +1,29 @@
...
+      "strategy": {
+        "type": "sequence"
+      },
+      "template": {
+        "metadata": {
+          "name": "string",
+          "labels": "object"
+        }
+      }
+    },
```
akihikokuroda, line 24:
```
"spec":	{
  "type": "object",
  "properties":	{
    "strategy": {
      "type": "object",
      "properties": {
        "type": {
          "type": "string",
          "description": "workflow strategy:, sequence, condition, parallel"
        }
      }
    }
    "template": {
      "type" "object",
      "properties": {
        "metadata": {
          "type": "object",
          "properties": {
            "name": {
              "type": "string",
              "description": "workflow name"
            },
            "labels": {
              "type": "object",
              "description": "workflow labels, key: value pairs"
            }
          }
        }
      }
    }
  }
}

```

test/schema/simple.yaml
planetf1: Agree. It also makes it easier to then add additional attributes for each step.

It also collapses the steps array (which defines the sequence) and the distinct step definitions in the original proposal, which feels as if it involves duplication

maximilien: OK. But my guess we will need both or a way to have steps separated at some point. But for now, since use cases are relatively simple, we can do this.

workflow_schema.json
```diff
@@ -0,0 +1,29 @@
...
+  "title": "Bee Hive Workflow Schema",
+  "description": "A JSON schema for defining Bee Hive workflows in YAML or JSON",
+  "type": "object",
+  "properties": {
+    "apiVersion": "string",
+    "kind": "string",
+    "metadata": {
+      "name": "string",
+      "labels": "object"
+    },
```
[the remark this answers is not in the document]
maximilien: Good one Aki. Thanks

workflow_schema.json
```diff
@@ -0,0 +1,29 @@
...
+          "name": "string",
+          "labels": "object"
+        }
+      }
+    },
+    "agents": [],
+    "exception": "string",
+    "steps": []
+  }
+}
```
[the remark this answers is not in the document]
maximilien: Yup. Will do as well.

maximilien:
See latest changes @akihikokuroda, @psschwei, and @planetf1 

Let's merge, if OK, since I have new version for conditionals

tools/workflow_schema.json
```diff
@@ -0,0 +1,41 @@
...
+        "description": "workflow name"
+      },
+      "labels": {
+        "type": "object",
+        "description": "workflow labels, key: value pairs"
+      }
+    },
+    "agents": [],
+    "exception": "string",
+    "steps": []
```
akihikokuroda, line 39: Please replace these lines with:
```
"agents": {
  "type": "array",
  "items": {
    "type": "string"
  }
}
"exception" {
  "type": "string"
}
"steps": {
  "type": "object"
}
```

akihikokuroda:
Other than the last comment.  It looks good to me.  Thanks!

tools/workflow_schema.json
maximilien: OK. Updating. Thx Aki

maximilien:
OK. Addressed. Resolving all issues.

maximilien:
OK. We are good for a merge here @akihikokuroda or others. Thx

tools/workflow_schema.json
```diff
@@ -0,0 +1,48 @@
...
+  },
+  "agents": {
+    "type": "array",
+    "items": {
+      "type": "string"
+    }
+  },
+  "steps": {
+    "type": "object"
+  },
```
psschwei, line 46: should this also be an array?

akihikokuroda: No, it should not.  It's a dictionary so just it's object.

akihikokuroda (approved):
LGTM!

tools/workflow_schema.json
maximilien: solved. Thx

PR #140, opened 2025-01-24
refactoring sequence method to allow for steps

### LINKED ISSUE #138: refactor workflow.py after steps PR update
in the previous PR, `step.py` was created. We need to refactor _sequence functions to have this step compatibility

developer-gliu:
fixes #138 

assuming schema looks something like this:
```
spec:
  strategy:
    type: sequence
    output: verbose
  template:
    prompt: "Initial prompt"
    steps:
      - name: step1
        agent: agent1
      - name: step2
        agent: agent2
      - name: step3
        agent: agent1
```
and we are passing all the output/prompt to the next step/agent

### DIFF
--- a/bee-hive/bee_hive/workflow.py
+++ b/bee-hive/bee_hive/workflow.py
@@ -38,15 +38,28 @@ class Workflow:
             print("not supported yet")   
 
     def _sequence(self):
+        # NOTE: "template" is an arbitrary key name that we use for now, it could be anything
         prompt = self.workflow["spec"]["template"]["prompt"]
-        for agent in self.agents.values():
-            if (
-                self.workflow["spec"]["strategy"]["output"]
-                and self.workflow["spec"]["strategy"]["output"] == "verbose"
-            ):
-                prompt = agent.run_streaming(prompt)
-            else:
-                prompt = agent.run(prompt)
+        steps = self.workflow["spec"]["template"]["steps"]
+
+        # Create a Step object for each step in the workflow
+        for step in steps:
+            if step["agent"]:
+                step["agent"] = self.agents.get(step["agent"])
+            self.steps[step["name"]] = Step(step)
+
+        # Run each step in sequence, assuming that the output of each step is the input to the next
+        for step_name in self.steps:
+            step = self.steps[step_name]
+            response = step.run(prompt)
+            prompt = response.get("prompt", prompt)
+
+        if (
+            self.workflow["spec"]["strategy"]["output"]
+            and self.workflow["spec"]["strategy"]["output"] == "verbose"
+        ):
+            print(f"Step {step_name} output: {response}")
+
         return prompt
 
     def _condition(self):


bee-hive/bee_hive/workflow.py
```diff
@@ -38,15 +38,28 @@ def run(self):
...
+
+        # Create a Step object for each step in the workflow
+        for step in steps:
+            if step["agent"]:
+                step["agent"] = self.agents.get(step["agent"])
+            self.steps[step["name"]] = Step(step)
+
+        # Run each step in sequence, assuming that the output of each step is the input to the next
+        for step_name in self.steps:
+            step = self.steps[step_name]
```
akihikokuroda, line 53: The `self.steps` has `step` objects instead of the step name.  So this should be simply
```
for  step in self.steps:
```

developer-gliu:
I'm having trouble deciding whether we should keep the verbosity feature inside of the sequence method, or keep it focused on the workflow orchestration. Seems like step.run already handles this, because we are choosing between the agent.run_streaming or simply agent.run (so if we were to add it, I think it should be in here instead?) Let me know your thoughts @akihikokuroda

akihikokuroda:
My previous review comments may not correct. I'm looking closely now.

bee-hive/bee_hive/workflow.py
```diff
@@ -39,14 +39,14 @@ def run(self):
...
-                and self.workflow["spec"]["strategy"]["output"] == "verbose"
-            ):
-                prompt = agent.run_streaming(prompt)
-            else:
-                prompt = agent.run(prompt)
+        steps = self.workflow["spec"]["template"]["steps"]
+        for step in steps:
+            if step["agent"]:
+                step["agent"] = self.agents.get(step["agent"])
+            self.steps[step["name"]] = Step(step)
```
akihikokuroda, line 46: I tried. this section should be like this.
```
        for step in steps:
            self.steps[step] = Step({"name": step, "agent": self.agents.get(step)} )
```

developer-gliu:
changed both the _sequence and _condition methods to instantiate the step class, changed _condition method to get the prompt and next steps correctly.

developer-gliu:
if this looks correct for now, Im going to write a test case to see if the behavior works for now @akihikokuroda

akihikokuroda:
The schema may change.  Please check [here](https://github.com/i-am-bee/bee-hive/pull/144#issuecomment-2613548785) Please revert the changes in `_condition` for now.

developer-gliu:
closing this issue for now, until schema is defined. @psschwei @akihikokuroda

maximilien:
> The schema may change. Please check [here](https://github.com/i-am-bee/bee-hive/pull/144#issuecomment-2613548785) Please revert the changes in `_condition` for now.

No not changing schema. Let's move with what we have now. We can change. shema later if there is a reason from one pf the use cases :)

PR #139, opened 2025-01-24
First version of agent schema and some refinement to workflow schema [fixes #136]

maximilien:
* new agent schema
* new tests agents in JSON and YML
* small changes to the workflow schema that fixes the new location .../be-hive/tools

CC: @developer-gliu @AngeloDanducci

PR #144, opened 2025-01-24
Fixes Issue #126: adding conditionals to workflow shema

### LINKED ISSUE #126: workflow: define schema with conditional
Build on issue #48 to add conditionals to steps

maximilien:
First version of schema with conditionals which would then fix #126. 

This includes ability to define conditionals on steps in both:

1. if/then/else
2. case/do/default 

Added some tests files in test/schema/conditional_workdlow.<yml/json>

The following section of the YAML test defines the parts that have changed:

```yaml
steps:
    - name: step1
      agent: agent1
      condition:
        - if: '{$in.some_condition == True}'
          then: step2
          else: step3
    - name: step2
      agent: agent2
    - name: step3
      agent: agent3
      condition:
        - case: '{$in.some_condition == True}'
          do: step2
        - case: '{$in.some_condition == False}'
          do: step3
        - default:
            do: step3
```

Please peruse and review @akihikokuroda @psschwei

Apologies that this PR also includes the changes for PR #139. I'll be careful next time to work on branches. I will rebase when you merge #139.

So for this one PLEASE only attention to files:

`tools/workflow_schema.json`
`test/schema/conditional_workflow.json`
`test/schema/conditional_workflow.yml`

CC: @planetf1 @developer-gliu @AngeloDanducci


test/schema/conditional_workflow.yml
```diff
@@ -0,0 +1,38 @@
...
+        use-case: test
+  agents:
+    - agent1
+    - agent2
+    - agent3
+  exception: step3
+  steps:
+    - name: step1
+      agent: agent1
+      condition:
```
akihikokuroda, line 24: Is this condition evaluated against the output of the agent1?

test/schema/conditional_workflow.yml
```diff
@@ -0,0 +1,38 @@
...
+      labels:
+        app: example
+        use-case: test
+  agents:
+    - agent1
+    - agent2
+    - agent3
+  exception: step3
+  steps:
+    - name: step1
```
akihikokuroda, line 22: Is the first step the step executed first when the workflow is executed?

test/schema/conditional_workflow.yml
```diff
@@ -0,0 +1,38 @@
...
+      agent: agent2
+    - name: step3
+      agent: agent3
+      condition:
+        - case: '{$in.some_condition == True}'
+          do: step2
+        - case: '{$in.some_condition == False}'
+          do: step3
+        - default:
+            do: step3
```
akihikokuroda, line 38: Does this mean the step3 executed repeatedly?  How is this end the workflow execution?

akihikokuroda (commented):
I have a few questions.  Thanks!

test/schema/conditional_workflow.yml
psschwei: default should be what happens if none of the case conditions are met. so say if it was `condition == 3 do X; condition == 4 do Y; default do Z` ... with true/false probably would just use if/then

test/schema/conditional_workflow.yml
akihikokuroda: How about adding `start` in the schema that specifies the start step name?

test/schema/conditional_workflow.yml
akihikokuroda: Can we assume that the condition expression is the python expression for now?

test/schema/conditional_workflow.yml
akihikokuroda: In addition to the step name, how about allowing `exit` or `end` so like `do:exit`?

akihikokuroda:
There are inconsistency between simple_workflow.yaml and conditional_workflow.yaml
in [simple_workflow.yaml](https://github.com/i-am-bee/bee-hive/blob/43f9e3eeed0c9f26f22eee34602c5b76c613d1f0/test/schema/simple_workflow.yaml#L21C1-L25C18)
```
  steps:
    step1: agent1
    step2: agent2
    step3: agent1
    step4: agent3
```
in conditional_workflow.yaml
```
  steps:
    - name: step1
      agent: agent1
      condition:xxx
    - name: step2
      agent: agent2
      condition:
      ....
```
I like the conditional way better.  Can we change the workflow schema to conditional way?

test/schema/conditional_workflow.yml
maximilien: Yes, it should be Python and can use `in` and `out` dictionaries

maximilien: > Is this condition evaluated against the output of the agent1?

Against the output of the agent from previous step, yes.

test/schema/conditional_workflow.yml
maximilien: > Does this mean the step3 executed repeatedly? How does this end the workflow execution?

yes :) I did this on purpose since we need to have iterations (for or while loop) in next iteration :)

maximilien: > In addition to the step name, how about allowing `exit` or `end` so like `do:exit`?

We will need to have loops.

maximilien:
> There are inconsistency between simple_workflow.yaml and conditional_workflow.yaml in [simple_workflow.yaml](https://github.com/i-am-bee/bee-hive/blob/43f9e3eeed0c9f26f22eee34602c5b76c613d1f0/test/schema/simple_workflow.yaml#L21C1-L25C18)
> 
> ```
>   steps:
>     step1: agent1
>     step2: agent2
>     step3: agent1
>     step4: agent3
> ```
> 
> in conditional_workflow.yaml
> 
> ```
>   steps:
>     - name: step1
>       agent: agent1
>       condition:xxx
>     - name: step2
>       agent: agent2
>       condition:
>       ....
> ```
> 
> I like the conditional way better. Can we change the workflow schema to conditional way?

Noted that as well since I think we need to have both. WDYT?

test/schema/conditional_workflow.yml
akihikokuroda: So is this workflow keep running?

test/schema/conditional_workflow.yml
akihikokuroda: When is the `agent1` in this step executed?

akihikokuroda:
2 different schema in the `steps` may confuse users.  I like to change the sequence one to 
```
steps:
  -name: step1
    agent: agent1
  -name: step2
    agent: agent2
```
to match the schema of condition.

test/schema/conditional_workflow.yml
maximilien: I like this. However, is the start not implicit from the first step specified?

test/schema/conditional_workflow.yml
maximilien: I’ll rewrite the example to allow it to end workflow but like any language you can have infinite loop.

maximilien:
> I like the conditional way better. Can we change the workflow schema to conditional way?

Can’t we have both? The first way when steps have no conditions? WDYT?

psschwei:
Won't the first way require hard coding `step1`, `step2`, etc. into the JSON schema?

psschwei:
(dropping the milestone on the PR as it's covered in the issue)

tools/workflow_schema.json
```diff
@@ -41,11 +41,40 @@
...
+              "do": "string",
+              "default": "string"
+            }
+          }
+        }
+      }
     }
-  },
-  "exception": "string"
+  }
```
akihikokuroda, line 79: I corrected the schema syntax.  Please verify this.
```
{
  "$schema": "https://json-schema.org/draft/2020-12/schema",
  "$id": "https://github.com/i-am-bee/bee-hive/tools/workflow.schema.json",
  "title": "BeeAI Hive",
  "description": "A schema for defining Bee Hive workflows in YAML or JSON",
  "type": "object",
  "properties": {
    "metadata": {
      "apiVersion": {
        "type": "string",
        "description": "beehive/v1"
      },
      "kind": {
        "type": "string",
        "description": "Workflow"
      },
      "type": "object",
      "properties": {
        "name": { 
          "type": "string",
          "description": "workflow name"
        },
        "labels": {
          "type": "object",
          "description": "workflow labels, key: value pairs"
        }
      },
      "required": [
        "name"
      ]
    },
    "spec": {
      "type": "object",
      "properties": {
        "strategy": {
          "type": "object",
          "properties": {
            "type": "string",
            "description": "sequence, condition"
          }
        },
        "template": {
          "type": "object",
          "properties": {
            "meatdata": {
              "type": "object",
              "properties": {
              "name": {
                "type": "string",
                "description": "workflow name"
              },
              "labels": {
                "type": "object",
                "description": "workflow labels, key: value pairs"
              }
            } 
          }  
        },
        "agents": {
          "type": "array",
          "items": {
            "type": "string"
          }
        },
        "exception": "string",
        "steps": {
          "type": "array",
          "items": {
            "type": "object",
            "properties": {
              "name": {
                "type": "string",
                "description": "step name"
              },
              "agent": {
                "type": "string",
                "description": "the agent for this step"
              },
              "condition": {
                "type": "array",
                "description": "if/then/else or case/do/default condition",
                "items": {
                  "type": "object",
                  "properties": {
                    "if": "string",
                    "then": "string",
                    "else": "string",
                    "case": "string",
                    "do": "string",
                    "default": "string"
                  }
                }
              }
            },
            "required": [
              "name",
              "agent"
            ]
          }
        }
      }
    }
  }
}

```

akihikokuroda: Here is a small python program that validate the yaml file against the json schema.
```
import sys
import yaml
import json
import jsonschema

def validate_yaml(yaml_file, schema_file):
    with open(yaml_file, 'r') as f:
        yaml_data = yaml.safe_load(f)
        json_data = json.dumps(yaml_data, indent=4)
        print(json_data)

    with open(schema_file, 'r') as f:
        schema = json.load(f)
        print(schema)

    jsonschema.validate(yaml_data, schema)
    print("YAML file is valid.")

if __name__ == "__main__":
    if len(sys.argv) != 3:
        print("Usage: python validate.py yamlfile schemafile")
    else:
        validate_yaml(sys.argv[1], sys.argv[2])
```

maximilien: Your schema does not validate Aki. Please try: https://jsonschema.dev/ and test schema:

```json
{
  "apiVersion": "beehive/v1",
  "kind": "Workflow",
  "metadata": {
    "name": "beehive-deployment",
    "labels": {
      "app": "example2"
    }
  },
  "spec": {
    "strategy": {
      "type": "sequence"
    },
    "template": {
      "metadata": {
        "name": "beehive-deployment",
        "labels": {
          "app": "example",
          "use-case": "test"
        }
      }
    },
    "agents": [
      "agent1",
      "agent2",
      "agent3"
    ],
    "exception": "step3",
    "steps": [
      {
        "name": "step1",
        "agent": "agent1",
        "condition": [
          {
            "if": "{$in.some_condition == True}",
            "then": "step2",
            "else": "step3"
          }
        ]
      },
      {
        "name": "step2",
        "agent": "agent2"
      },
      {
        "name": "step3",
        "agent": "agent3",
        "condition": [
          {
            "case": "{$in.some_condition == True}",
            "do": "step2"
          },
          {
            "case": "{$in.some_condition == False}",
            "do": "step3"
          },
          {
            "default": {
              "do": "step3"
            }
          }
        ]
      }
    ]
  }
}
```

maximilien:
> Won't the first way require hard coding `step1`, `step2`, etc. into the JSON schema? (which would mean users could only run the number of steps we've defined, i.e. if we go up to `step5`, users couldn't run a 6 step workflow)

OK what about this compromise @akihikokuroda and @psschwei. The first approach and second should not be mixed. The first is for simple workflows with sequential steps and the second adds more complexity. We can work on either only having one approach or mix and match in future.

If you are all OK then let's merge this PR.

test/schema/conditional_workflow.yml
maximilien: Steps are executed in the order they are defined which is natural. No need for start but noted suggestion. Let see how far we can get...

tools/workflow_schema.json
akihikokuroda: @maximilien I tried this with the schema https://github.com/i-am-bee/bee-hive/pull/144#discussion_r1931056383 in the online validator.  It validated fine.  I also tried test/schema/conditional_workflow.json, too. It validated, too.

planetf1:
I checked the code 

If I take the code from 72cdbdd27c18bd812f21a770496b718653cd551d &
 - convert yaml workflow to json ([web tool](https://www.bairesdev.com/tools/json2yaml/))
 - check yaml schema against workflow ([web tool](https://www.jsonschemavalidator.net/)
 
 I get a failure, but it's an object to the `"$schema": "https://json-schema.org/draft/2020-12/schema",` line. This seems to cause some tools an issue (such as pajv) but looks ok according to the current [spec](https://json-schema.org/draft/2020-12)?
 
 Removing this passes.

planetf1:
I think we should definitely only to only have a single schema, or we have a different kind of resource. We also can have optional elements. I do think we should capture 'best so far' and merge, and then persue via a subsequent issue/pr

akihikokuroda:
> convert yaml workflow to json ([web tool](https://www.bairesdev.com/tools/json2yaml/))
> check yaml schema against workflow ([web tool](https://www.jsonschemavalidator.net/)

FYI @planetf1 I put the python program to do these in bee-hive/test/schema/validator.py

tools/workflow_schema.json
maximilien: Corrected schema works, let me update so we can close this issue.

maximilien:
OK @akihikokuroda please review and merge if OK. I'll monitor if there are any more changes required. Thx

akihikokuroda (approved):
LGTM!  Thanks!

PR #150, opened 2025-01-28
New schema refactor

developer-gliu:
Refactoring sequence method only to match workflow.json schema. I didn't fix conditional method at all yet, just want to someone else to take a look and see that it makes sense so far.

Note: I've added a test in the middle of bee-hive file just to test the sequence method quickly and not have to deal with all the circular imports (but I will fix that later)


psschwei:
looks good to me so far

maximilien (commented):
Love this. Thanks for adding a small test @developer-gliu 

Let me try to run it. And will build on it. Thx

bee-hive/bee_hive/test_workflow.py
```diff
@@ -0,0 +1,42 @@
+import unittest
```
planetf1, line 1: if you put this under bee_hive/tests it would be consistent with the other test previously added, managed by pytest ? It would work even in current location, but keeps the tree tidier

(ie `poetry run pytest`)

planetf1 (changes requested):

bee-hive/bee_hive/test_workflow.py
planetf1: You'll need to rebase on the latest main to get the pytest setup

bee-hive/bee_hive/test_workflow.py
```diff
@@ -0,0 +1,42 @@
...
+from unittest.mock import MagicMock
+from workflow import Workflow
```
planetf1, line 3:
```suggestion
from bee_hive.workflow import Workflow
```

With this change, the test should run under pytest (whatever directory it is in) -- as the module is well defined
Without it it's very dependent on path

planetf1:
Get:
```
FAILED tests/crewai/test_crewai.py::CrewAITest::test_agent_runs - KeyError: 'prompt'
```
When rebased on main. Will take a look.

bee-hive/bee_hive/workflow.py
```diff
@@ -39,15 +39,19 @@ def run(self):
...
+            step_name = step["name"]
+            agent_name = step["agent"]
+            agent_instance = self.agents.get(agent_name)
+            if not agent_instance:
+                raise ValueError(f"Agent {agent_name} not found for step {step_name}")
+            self.steps[step_name] = Step({"name": step_name, "agent": agent_instance})
+
+        for step_name, step_obj in self.steps.items():
+            response = step_obj.run(prompt)  
+            prompt = response.get("prompt", prompt)
```
planetf1, line 54: prompt is currently not required by spec - so either this should be optional, or spec should be update to state as required

planetf1 (changes requested):

planetf1:
The breakage is down to changes in `workflow.py`. The relevant workflow run by the failing test is in `tests/crewai/workflow.yml`.

Previously the code was getting the prompt from `spec.template.prompt`
The updated code is getting it from `spec.prompt`

Which is correct, or rather the intention for this PR.
The [current schema](https://github.com/i-am-bee/bee-hive/blob/main/tools/workflow_schema.json) in main is using the second of these. So I would expect updating the workflow to drop the template would help

However additionally that schema does not **require** a prompt (only agent and name). If this is the case, the workflow should be tolerant of that.

To get the test working, update the tests/crewai/workflow.yml to be
```
apiVersion: beehive/v1alpha1
kind: Workflow
metadata:
  name: beehive-deployment
  labels:
    app: mas-example
spec:
  strategy:
    type: sequence
  agents:
    - name: tests.crewai.crew_dummy.DummyCrew.dummy_crew
  prompt: Nothing
  start: begin
  steps:
    - name: begin
      agent: tests.crewai.crew_dummy.DummyCrew.dummy_crew
      prompt: Welcome
      dependencies:

```
optimise as you see fit.

In my environment (so based on main) that works fine.
I can't change it with the earlier workflow.yaml (unless I duplicated under *template*)

The obscure agent name will be addressed in a future update.

It's really good we're adding yet more tests, a spec, and validation as it will make evolution easier in future!

bee-hive/bee_hive/workflow.py
planetf1: Ok if we want to defer, as really we need full spec validation, or much more defensive code for parsing the objects.

bee-hive/bee_hive/test_workflow.py
planetf1: Not critical, but this now means we have test code in
test/ 
bee-hive/bee_hive < your test
bee-hive/tests < tests I added managed by poetry/pytest
test (schema examples)

I think separating tests is probably clearer. Also for automated unit tests, having it under poetry makes sense, so that would be my preference

planetf1:
A similar update will be needed in bee test from #153, depending on order of merge

bee-hive/bee_hive/workflow.py
```diff
@@ -60,15 +60,21 @@ def run(self):
...
-        for agent in self.agents.values():
-            if (
-                self.workflow["spec"]["strategy"]["output"]
-                and self.workflow["spec"]["strategy"]["output"] == "verbose"
-            ):
-                prompt = agent.run_streaming(prompt)
-            else:
-                prompt = agent.run(prompt)
+        prompt = self.workflow["spec"].get("prompt","")   # using get b/c prompt is optional
+        steps = self.workflow["spec"].get("steps",[])
```
psschwei, line 64: Steps are in `spec.template`, so I think you may need to do

```suggestion
        steps = self.workflow["spec"]["template"].get("steps",[])
```

(syntax may not be exactly correct)

bee-hive/tests/workflow/test_sequence.py
```diff
@@ -0,0 +1,40 @@
...
+class TestSequenceMethod(unittest.TestCase):
+    def setUp(self):
+        self.mock_workflow = {
+            "spec": {
+                "prompt": "Start of the workflow",
+                "steps": [
+                    {"name": "step1", "agent": "agent1"},
+                    {"name": "step2", "agent": "agent2"},
+                    {"name": "step3", "agent": "agent1"}
+                ]
```
psschwei, line 14: I think steps here would also need to move under `spec.template` instead of just `spec`

bee-hive/bee_hive/workflow.py
developer-gliu: How come there should be template? I'm basing it off the latest workflow_schema merged, and if we look at that the organization is like this:
specs
    template
    agents
    prompt
    steps

so from what I see, it doesnt seem necessary

psschwei: you're right, we updated the schema but not the test

psschwei:
#157 will fix the failing test

developer-gliu:
currently when I'm running the test, my output looks like this:
{'prompt': '{\'prompt\': "{\'prompt\': \'Start of the workflow processed by agent1\'} processed by agent2"} processed by agent1'}

This is the correct output, but it has a triple nest because of how we designed the sequence method. This gives us an easy way to access and see what each prompt is if we look back at it, but is a bit complex.

On the other hand, I could flatten it if we only care about the final prompt result. This would make it cleaner. 

Which way should we go with since I'm not too sure on the end goal

maximilien (commented):
I like this. The one one thing I would do is to verify that all the agents are executing in certain order. So your mock agent can just output their name or number.

Then your test should verify the sequence. So for a simple workflow of three agents 0 to 2 then you can capture the output of running the workflow and test that you see the sequence you expect, e.g., Mock0Mock1Mock2

For conditional just have each agent as they are executing in a step set a boolean flag in the output dictionary, e.g., `out.<AgentName>=True` which will then be copied as the `input` and `in` alias for the next step. 

Then in the workflow you can use that value to check that conditions are executing or not.

Hope this helps :)

bee-hive/poetry.lock
```diff
@@ -0,0 +1,600 @@
+# This file is automatically @generated by Poetry 2.0.1 and should not be changed by hand.
```
planetf1, line 1: this file should be removed from the PR (it can help with reproducibility, but causes additional overhead I don't think we need currently - as per team discussion)

planetf1 (dismissed):

developer-gliu:
added execution order test as suggested by max. ready for review now

planetf1 (approved):

PR #153, opened 2025-01-29
feat(test): Add simple Bee Agent framework test

planetf1:
* Adds a trivial test for a Bee Agent based workflow using mock
* Removes `poetry.lock` & adds to .gitignore

The BeeAgent class is mocked & the run() method returns 'OK' concatenated with the input prompt from the single-step workflow

This validates the parsing of the workflow & ensures the crewai vs beeai agent check is correct.

poetry.lock is removed as otherwise anytime there are significant changes, `poetry install` will fail. This can be annoying at the tty, but also breaks automated tests. They could be corrected, however whether to include a lock file or not is debatable. The value is in reproducible builds, but in the early stages of development it seems easier to do without. See [poetry docs](https://python-poetry.org/docs/basic-usage/#committing-your-poetrylock-file-to-version-control)

**Background:**
 - We need to consider how we structure our tests, and specifically need to mock out a variety of classes in many cases (though we also want integration tests that use an LLM). This is a first-level mock. later tests of an actual bee agent may just mock out openAPI for example.
  - this brings parity for bee agents to the crewAI agent test (effectively)
  - helps to avoid regressions
  
**Caveats:**
  -  The implementation of BeeAgent (including definitions in agent.yaml) is not tested 
  - only a single stage simple workflow is tested


planetf1:
Sample output:
```
(.venv) ➜  bee-hive git:(issue119f) poetry run pytest                                                 
============================================== test session starts ===============================================
platform darwin -- Python 3.12.8, pytest-8.3.4, pluggy-1.5.0 -- /Users/jonesn/AI/bee/bee-hive/.venv/bin/python3.12
cachedir: .pytest_cache
rootdir: /Users/jonesn/AI/bee/bee-hive/bee-hive
configfile: pyproject.toml
plugins: anyio-4.8.0, mock-3.14.0
collected 2 items                                                                                                

tests/bee/test_bee.py::test_agent_runs OK:Welcome
PASSED
tests/crewai/test_crewai.py::CrewAITest::test_agent_runs Running CrewAI agent: tests.crewai.crew_dummy.DummyCrew.dummy_crew with prompt: OK:Nothing
Getting a Crew to return
Running kickoff method
OK:Nothing
OK
PASSED

=============================================== 2 passed in 0.49s ================================================

```

bee-hive/tests/bee/workflow.yaml
```diff
@@ -0,0 +1,22 @@
...
+spec:
+  strategy:
+    type: sequence
+    output: nonal
+  template:
+    metadata:
+      labels:
+        app: testapp
+    agents:
+      - name: beetest
```
akihikokuroda, line 16: `agents` lists the agent used in the steps.  So it should be `crewtest`.

bee-hive/tests/bee/workflow.yaml
```diff
@@ -0,0 +1,22 @@
...
+    prompt: Welcome
+    start: begin
```
akihikokuroda, line 18: The `start` is not in the schema for now.

bee-hive/tests/bee/workflow.yaml
```diff
@@ -0,0 +1,22 @@
...
+    steps:
+      - name: begin
+        agent: crewtest
+        dependencies:
```
akihikokuroda, line 22: The `dependencies` is not in the schema for now.

akihikokuroda (changes requested):
Looks good.  Thanks!  I have a few minor comments.

bee-hive/tests/bee/workflow.yaml
planetf1: removed from bee & crewai test

bee-hive/tests/bee/workflow.yaml
planetf1: removed from both tests

bee-hive/tests/bee/workflow.yaml
planetf1: fixed in both tests

planetf1:
thanks for the comments @akihikokuroda - they should be addressed now

PR #154, opened 2025-01-29
fix(internal): fix schema format (spaces/tabs) 

planetf1:
Fixes the schema to use spaces (not tabs)

I was getting visual layout issues with the schema in vscode.

For example

Screenshot 2025-01-29 at 11 47 48

This fix replaces with standard spaces

### DIFF
--- a/tools/workflow_schema.json
+++ b/tools/workflow_schema.json
@@ -25,9 +25,7 @@
           "description": "workflow labels, key: value pairs"
         }
       },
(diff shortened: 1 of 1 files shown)


psschwei (approved):

maximilien:
I have not tried this but this GH Action seems to be a solution:

https://github.com/GrantBirki/json-yaml-validate

PR #162, opened 2025-01-30
move bee_hive to root level

psschwei:
bee-py code has all moved to its own repo, so let's remove it from here and move bee_hive to main level


maximilien (approved):
Love it. Thx

tests/examples/test_condition.py
```diff
@@ -8,7 +8,7 @@
 import dotenv
 from openai import OpenAI
 import yaml
-from workflow import Workflow
+from bee_hive.workflow import Workflow
```
planetf1, line 11: I'm not sure if this is appropriate - it depends on how we want to manage imports

If we want relative imports within the package, we may want to begin with . or ..
but if we qualify by module (bee_hive), it seems as if the code will work in more situations - this is why the test code currently uses the same construct. For example tests/bee/test_bee.py . It also will be consistent with external use.

The directory moving isn't relevant here, since the module name has remained constant, rather than being a path

planetf1 (changes requested):

planetf1:
See comment above ref test code, otherwise looks good

tests/examples/test_condition.py
psschwei: >  this is why the test code currently uses the same construct

I'd keep it consistent for now... if down the line it causes problems we can always refactor

maximilien: I assume we will plan this big change after POC1 demos?

psschwei: that and once the bee-py repo goes public

psschwei:
will revisit this after we make more moves into this repo

PR #172, opened 2025-02-04
Add MockAgent

akihikokuroda:
MockAgent is added for testing.

MockAgent is created instead of BeeAgent or CrewAI agent when "BEE_TEST" env variable is set like `export BEE_TEST=1`


psschwei (approved):

maximilien (commented):
GREAT, except can we change BEE_TEST to DRY_RUN since this would reflect more what we are trying to do. Especially since this project is independent of agents, e.g., Bee or Crew or LangChain etc.

Thanks for doing this. And please merge right after addressing above and I will start using it ASAP. Thx again.

PR #184, opened 2025-02-05
Separate agent creation from workflow creation

akihikokuroda:
The agent can be create (by constructor) and saved (by save_agent method in agent.py)
The workflow constructor can take an array of agent definitions, an array of saved agent names or `None`(this case the agents listed in the workflow must be created and saved).

```python
import json
import os
import sys

import dotenv
import yaml
from bee_hive.workflow import Workflow
from bee_hive.bee_agent import BeeAgent
from bee_hive.agent import save_agent

dotenv.load_dotenv()

def parse_yaml(file_path):
    return yaml_data

if __name__ == "__main__":
    agents_yaml = parse_yaml("bee_hive/joke_agents_new.yaml")
    for agent_def in agents_yaml:
        agent = BeeAgent(agent_def)
        save_agent(agent)
    
    workflow_yaml = parse_yaml("bee_hive/joke_workflow_new.yaml")
    try:
        workflow = Workflow(None, workflow_yaml[0])
    except Exception as excep:
        raise RuntimeError("Unable to create agents") from except
    workflow.run()

```

### DIFF
--- a/bee-hive/bee_hive/agent.py
+++ b/bee-hive/bee_hive/agent.py
@@ -2,6 +2,8 @@
 # SPDX-License-Identifier: Apache-2.0
 
 from abc import abstractmethod
+import os
+import pickle
 
 class Agent:
     """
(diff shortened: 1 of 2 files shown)


psschwei:
It looks like you're storing the BeeAgent python object in this code, right?

```python
for agent_def in agents_yaml:
        agent = BeeAgent(agent_def)
        save_agent(agent)
```

What do you think about just saving the YAML instead of the BeeAgent object? So doing something like  `save_agent(agent_def)`.  In theory, that could just be written out as JSON to `agent.db`, which would eliminate the need for pickling.

akihikokuroda:
@psschwei The key is that the agent (especially Bee agent) is created so that the agent can be tweaked before the workflow run.

psschwei:
When you say created, do you mean in the bee-stack or the BeeAgent python object? If the former would

```python
for agent_def in agents_yaml:
        agent = BeeAgent(agent_def)
        save_agent(agent_def)
```

be sufficient? If we needed to reload the agent, then possibly something like this could work 

```python
def restore_agent(agent_name: str) -> Agent:
    """
    Restore agent from storage.
    """
    agents = _load_agent_db()
    agent_data= agents[agent_name]
    agent = BeeAgent(agent_data)
    return(agent)
```

I've seen some subtle issues due to pickle in the past, as well as some security vulnerabilities, so I try to avoid it if possible. But I'm also just thinking out loud, so feel free to proceed how you see best.

akihikokuroda:
What I mean by `created` is agent in bee-stack created so that the agent is tweaked in the BeeAgent UI.  The saved contents are the same information in the agent definition yaml and some id that can be used to identify the remote agent. (Agent in BeStack, CrewAi agent  and etc..)

psschwei (approved):

PR #190, opened 2025-02-06
refactor + implementing comments from demo

george-lhj:
put all the code into a sequential folder since I will change the next version to be conditional. Also implementing some naming changes based upon feedback in the demo

will probably make files in this more organized later, but I think should just merge and work on conditional workflow first


psschwei (commented):
In the future, the README instructions for running are a bit unclear. I'd pick either python or hive as the way to run and drop the other one (eventually, we'll have one CLI so may be best to just wait until then?)

psschwei:
I also didn't see a filtering tool in the UI

psschwei (approved):
can fix the tool issue later, let's get this merged for now so we can restructure tomorrow

george-lhj:
I refactored this already in pr #194 , so closing this without merging.

PR #171, opened 2025-02-04
test: added tests for `run`, `deploy`, and `validate` CLI commands

maximilien:
This is still WIP because:

- [x] need to have option for `--dry-run` using Mocks
- [x] missing create tests (need PR #169 to merge)
- [x] wire the `create` command
- [x] pass tests
- [x] wire remaining `run` and `deploy`  commands (moved to a TODO)

CC: @planetf1 and @psschwei take a look or wait for next update

PR #192, opened 2025-02-06
fix mock agent and make simple yamls consistent

akihikokuroda:
2 changes for testing

1. Fix bug in MockAgent.
2. make `test/yamls/agents/simple_agent.yaml` and `test/yamls/workflows/simple_workflow.yaml` consistent


akihikokuroda:
Fix for #171.

akihikokuroda:
I tested with #171.  It still fails the test but it may be fixed in the cli side.

I run this command with the #171 changes:
```
python -m unittest discover cli
```

psschwei (approved):

PR #194, opened 2025-02-06
refactor +impelementing demo comments pt 2.

george-lhj:
extension of pr#120

I refactored the demo to match the layout and execution of weather.ai example completely. Now, we use the same steps to execute the workflow.


bee-hive-demos/demos/summary.ai/agent_store.json
```diff
@@ -0,0 +1 @@
+{"Search Arxiv": "asst_67a514988a306469596fffa6", "Intermediary": "asst_67a514988a306469596fffa8", "Filter": "asst_67a514988a306469596fffaa", "Summary": "asst_67a514988a306469596fffac"}
```
psschwei, line 1: you might want to commit this as an empty dict and then later (after this PR merges) add it to the gitignore so new changes don't keep getting pushed

```suggestion
{}
```

bee-hive-demos/demos/summary.ai/workflow.yaml
```diff
@@ -1,33 +1,58 @@
...
+#         agent: Intermediary
+#         condition:
+#           - default:
+#               do: Filter
+#       - name: Filter
+#         agent: Filter
+#         condition:
+#           - default:
+#               do: Summary
+#       - name: end
```
psschwei, line 58:
```suggestion
```

psschwei (approved):
demo runs, so just the few things above and then we can merge

AngeloDanducci (approved):

bee-hive-demos/demos/summary.ai/agent_store.json
AngeloDanducci: Good point here, I think this also happens with the .env file if but I'll double check the gitignore

Issue #178, opened 2025-02-05
Inconsistency between spec validation & workflow execution


planetf1:
Testing in bee-hive-demos
```console
(bee-hive-demos-py3.12) ➜  bee-hive-demos git:(dev/jonesn/activitymerge) /Users/jonesn/AI/bee/bee-hive-demos/.venv/bin/python /Users/jonesn/AI/bee/bee-hive-demos/demos/activity-planner.ai/src/run.py
Enable the weather tool in the Bee UI
Enable the weather tool in the Bee UI
Enable the web_search tool in the Bee UI
Enable the web_search tool in the Bee UI
Traceback (most recent call last):
  File "/Users/jonesn/AI/bee/bee-hive-demos/demos/activity-planner.ai/src/run.py", line 32, in <module>
    test_agent_runs()
  File "/Users/jonesn/AI/bee/bee-hive-demos/demos/activity-planner.ai/src/run.py", line 28, in test_agent_runs
    result = workflow.run()
             ^^^^^^^^^^^^^^
  File "/Users/jonesn/AI/bee/bee-hive-demos/.venv/lib/python3.12/site-packages/bee_hive/workflow.py", line 61, in run
    return self._condition()
           ^^^^^^^^^^^^^^^^^
  File "/Users/jonesn/AI/bee/bee-hive-demos/.venv/lib/python3.12/site-packages/bee_hive/workflow.py", line 83, in _condition
    prompt = self.workflow["spec"]["prompt"]
             ~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^
KeyError: 'prompt'
```
occurs given a workflow:
```yaml
apiVersion: beehive/v1alpha1
kind: Workflow
metadata:
  name: beehive-deployment
  labels:
    app: mas-example
spec:
  strategy:
    type: condition
    output: nonal
  template:
    metadata:
      labels:
        app: mas-example
    agents:
      - name: current-affairs
      - name: hot-or-not
      - name: cold-activities
      - name: hot-activities
    prompt: New York City
    steps:
      - name: current-affairs
        agent: current-affairs
        condition:
        - default:
          do: hot-or-not
      - name: hot-or-not
        agent: hot-or-not
        condition:
        - if: (input.find('hotter') != -1)
          do: cold-activities
          else: hot-activities
      - name: end
```
which passes schema validation ok

planetf1:
[...]

PR #196, opened 2025-02-07
feat(internal): Add schema check to ci

planetf1:
Adds CI check for schema validation ([CLICK FOR EXAMPLE RUN](https://github.com/i-am-bee/bee-hive/actions/runs/13197461780))

Notes:
* we don't currently enforce tests pass via a branch protection rule - in fact we can't in a private repo
* Added as separate task to make it clear this is a different check (as we get set up)
* The crewai agent also uses a default name 'agents.yaml' for it's own agent definition! This will be renamed in a future PR

Partially addresses #178

### DIFF
--- a/tools/ci/check-schemas.sh
+++ b/tools/ci/check-schemas.sh
@@ -1,0 +1,51 @@
+#!/usr/bin/env bash
+
+# Runs validator, and composes markdown summary for github action
+
+if [ -z "$GITHUB_STEP_SUMMARY" ]
+then
+    GITHUB_STEP_SUMMARY=$(mktemp)
+fi
+
+declare -i fail=0
+WORKFLOW_FILES=$(find . -name '*workflow*.yaml')
+AGENT_FILES=$(find . -name '*agents*.yaml')
+
+
+echo "|Filename|Type|Stats|" >> "$GITHUB_STEP_SUMMARY"
+echo "|---|---|---|" >> "$GITHUB_STEP_SUMMARY"
+
+
+# Check workflows
+# TODO Consolidate duplication
+for f in $WORKFLOW_FILES
+do
+    if ! beeAI validate tools/workflow_schema.json "$f"
+    then
+      RESULT="FAIL ❌"
+      fail+=1
+    else
+      RESULT="PASS ✅"
+    fi
+    echo "|$f|workflow|$RESULT|" >> "$GITHUB_STEP_SUMMARY"
+done
+
+# Check agents
+for f in $AGENT_FILES
+do
+    if ! beeAI validate tools/agent_schema.json "$f"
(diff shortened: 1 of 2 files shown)


planetf1:
Output consists of the full job output, plus a summary (example below)

Screenshot 2025-02-07 at 10 01 29

planetf1:
This is a very basic check. A few other ideas
 - we could reference the schema in the yaml rather than depend on filenames

AngeloDanducci (approved):
This is great, though we may need to update some of the yaml files since the schema change PR is in.

Two notes
1. We can enforce this passes before merge since we're in a public repository.
2.  I think it's fine to scan all agent / workflow definitions.  Longer term we should think about how we would handle an agent.yaml file with a different agent type (bee vs crewai etc) or multiple types of agents in the same file.

tools/ci/check-schemas.sh
```diff
@@ -0,0 +1,51 @@
...
+# Runs validator, and composes markdown summary for github action
+
+if [ -z "$GITHUB_STEP_SUMMARY" ]
+then
+    GITHUB_STEP_SUMMARY=$(mktemp)
+fi
+
+declare -i fail=0
+WORKFLOW_FILES=$(find . -name '*workflow*.yaml')
+AGENT_FILES=$(find . -name '*agents*.yaml')
```
psschwei, line 12: question: would it make sense to be able to specify the files in the action? 

(we can add that later if we want, don't want to spent time on it right now)

psschwei (approved):

maximilien (approved):
LGTM. Address Paul's comment and then let's merge. I am OK with CI fails that we resolve later. Thx for this work @planetf1

tools/ci/check-schemas.sh
planetf1: Yes it would. This was a quick & dirty initial pass -- since I already was running a script locally in any case.
@AngeloDanducci 
- eventually it should be enforced (as you mentioned, public here now) - once we've fixed the backlog - which is why I think getting the change in is helpful
- The scan should not depend on filename. My preference would be to include a schema ref in the yaml. However this isn't standard. The [Red Hat yaml plugin for vscode](https://github.com/redhat-developer/vscode-yaml#associating-a-schema-in-the-yaml-file) uses `# yaml-language-server: $schema=<urlToTheSchema>`. Since these are our files, we could respect this as a hint, and whilst not requiring it to work
- If it is a workflow of ours, the bee vs crew should be determined by the schema. Either a single schema that's flexible enough (I think preferable) or two different schemas, with a different resource type.

-> will create issue

planetf1: See #205. Closing thread

PR #201, opened 2025-02-07
fix readme after rename

psschwei:


psschwei:
replaced by #202

maximilien (approved):
I did a similar change too but yours is better.

Let's also move all folders and files that are not generic "labs" into maestro

PR #204, opened 2025-02-07
Update tests.yaml

maximilien:
cd to maestro since all now relative to that directory


maximilien:
Actually should fix this in my PR #171 so closing this

Issue #217, opened 2025-02-12
Fix CI issues after restructure


maximilien:
@akihikokuroda let's use this issue to get all CI and tests passing.

I am running tests as follows:

```bash
(maestro-py3.12) (base) ➜  beeai-labs git:(main) pwd
/Users/maximilien/github/beeai-labs
(maestro-py3.12) (base) ➜  beeai-labs git:(main) pytest
======================================== test session starts ========================================
platform darwin -- Python 3.12.8, pytest-8.3.4, pluggy-1.5.0
rootdir: /Users/maximilien/github/beeai-labs
plugins: anyio-4.8.0, mock-3.14.0
collected 0 items / 5 errors

============================================== ERRORS ===============================================
...
```

Full `pytest` output: [pytest-output-02-12-25.txt](https://github.com/user-attachments/files/18773708/pytest-output-02-12-25.txt)

CC: @george-lhj

PR #218, opened 2025-02-12
Fix CI test failures caused by the restructure

akihikokuroda:
Quick fix for #217


maximilien (approved):
Lgtm

maximilien:
I think I fixed that too but ok merge should be noop

Issue #213, opened 2025-02-11
Complete `run`, `create`, and `deploy` CLI commands


maximilien:
- [x] ensure `--dry-run` tests passing
- [x] complete `Run` and `Create` CLI classes with command
- [x] add simple integration: validate, create, run, and deploy in `--dry-run`
- [ ] complete `Deploy` CLI classes with command

maximilien:
Working on this now @george-lhj 

Feel free to comment here what's missing so I can ensure I address for you. Thx.

george-lhj:
As a go deeper into the demo and figure whats missing I will update you, for now I think its okay

maximilien:
@akihikokuroda I don't see how I can do a deploy. Do you have an example or `Deploy` class? LMK when you have a minute --- not urgent but since #222 should use that LMK so I can have CLI code ready for you. Thx

maximilien:
One example:

```bash
$maestro deploy —docker xxx agents.yaml workflow.yaml
…
…
Workflow running on: https://localhost:4444
```

maximilien:
Closing since last sub task is now issue #235

PR #230, opened 2025-02-14
Fixes parts of issue #213

maximilien:
* refactor `Workflow` class to allow `create_agents()` separately
* re-organized `Workflow` class private and static methods
* implemented CLI `create` command to use new `Workflow::create_agents()`
* removed `tools/run_workflow.py` since no longer needed (use CLI `run` command instead)
* various small clean ups

CC: @george-lhj


akihikokuroda:
The workflow was updated to use the saved agents if the agents.yaml is not given or it is a string array of the agent names.  The agent can be saved by save_agent method in agent.py.  I was thinking that the agents are created and saved in the cli and the workflow restores and uses them but it's good to expose the create_agent method from the workflow.py to do the agent creation.

akihikokuroda (approved):

PR #231, opened 2025-02-17
save agent between create agent and run workflow

akihikokuroda:
The agents are created and saved by
```
./maestro create agent.yaml
```
The created agents are restored and used in the workflow run by
```
./maestro run None workflow.yaml
```
The "None" is the constant keyword for using the saved agents in the workflow run.

### DIFF
--- a/maestro/cli/commands.py
+++ b/maestro/cli/commands.py
@@ -17,7 +17,7 @@ import os, yaml, json, jsonschema, traceback
 from openai import OpenAI
 from jsonschema.exceptions import ValidationError, SchemaError
 
-from src.workflow import Workflow
+from src.workflow import Workflow, create_agents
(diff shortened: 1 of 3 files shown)


maximilien (changes requested):
Love it @akihikokuroda and thanks for adding. My only comment is to add one or two new tests in `tests/cli/test_commands.py` to check the new behavior, e.g., running with None. OK to do it with `--dry-run` so you can just create a new `test_create_agents_xyz(self):` and `test_run_agents(self):` methods (or whatever name you want). Once you add I will approve for merge. Thanks again.

maximilien (approved):
Love it. Merging. But likely will need to resolve 232 tomorrow. Feel free to start reviewing @akihikokuroda otherwise tomorrow. Cheers.

PR #233, opened 2025-02-18
Add input step

### LINKED ISSUE #223: maestro run pause workflow ability
In the demos, we often need tools inside of our agents. Right now, the command `maestro run` willl just execute the yaml file without any stopping. However, after creating the agents, we need to be able to pause and enable these tools (especially custom tools) in the agents before continuing to run.
Example, need to enable the 2 tools listed here:
Image

Additionally, some of the demos require user input. For instance, after retrieving k amount of research papers, lets say the user wants to select 1,2,7 indexed papers, then we need the ability to pause and also interact. 

These might be separate issues, but same theme in these 2 problems is to have the ability to pause.

akihikokuroda:
fix #223 

Add `input` in the step.  It does:  

1. shows an input prompt and wait for user input
2.  generate the next prompt by replacing the `{prompt}` and `{response}`. in the template

An example of the input step is:
```
      - name: input
        input:
          prompt: "Guess the number 1 through 6: "
          template: Is the number {prompt} is same as {response}?  
```

### DIFF
--- a/maestro/schemas/workflow_schema.json
+++ b/maestro/schemas/workflow_schema.json
@@ -95,7 +95,7 @@
                           "description": "template string for next prompt, {prompt} and {respomse} are replaced by the previos prompt and the user response"
                       }
                     }
-                  }
+                  },
                   "condition": {
                     "type": "array",
                     "description": "if/then/else or case/do/default condition",
@@ -125,8 +125,7 @@
                   }
                 },
                 "required": [
-                    "name",
-                    "agent"
+                    "name"
                 ]
               }
             }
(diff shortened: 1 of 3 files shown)


george-lhj:
thanks, this solves one of my issues

george-lhj (approved):

george-lhj:
please take a look at #234 to lmk your thoughts as it is related to this one

maestro/schemas/workflow_schema.json
```diff
@@ -111,8 +125,7 @@
                   }
                 },
                 "required": [
-                    "name",
-                    "agent"
+                    "name"
```
maximilien, line 128: why is `agent` removed from required?

maximilien (changes requested):
Looks good. LMK about my one question? If was by mistake then please include again, otherwise LMK. Otherwise, LGTM

PR #234, opened 2025-02-18
Refactoring demo after schema changes and user interaction.

george-lhj:
Updated read to add instructions, changed schema to allow for user interaction in choosing summaries, and updated demo to work w/ newest version after refactors.


maximilien:
What about having a loop feature instead so something like:

```yaml
...
    steps:
      - name: step1
        agent: test1
       while: (len(input.list) == 0)
        - do: step2
      - name: step2
        agent: test2
...

```

Of course we could also do it like this:

```yaml
...
    steps:
      - name: step1
        agent: test1
        loop:
        - until: (len(input.list) == 0)
        - do: step2
      - name: step2
        agent: test2
...
```

Preference? 1 or 2?

CC: @akihikokuroda WDYT?

maximilien (changes requested):
See my proposal. Vote or propose new approach and see if you could use it @george-lhj

george-lhj:
I think the loop/until format is easier to extend for parallel execution and possibly having loop limits, so I would vote for that one

maximilien:
Let me create a separate issue for loop in workflow and reference https://github.com/i-am-bee/beeai-labs/pull/234#issuecomment-2666884610 and propose the appropriate workflow schema changes. And we can from there on implementation.

@george-lhj please see if you can write your workflow with your choice and share here or in issue # <new_issue>

george-lhj:
finished refactored version, want to update main before going deeper into this demo and adding loops, new tools, etc

maximilien (changes requested):
Please add a `run.sh` and `test.sh` to run and test workflow using CLI. For test use `--dry-run` so no mock agents are used.

maximilien:
If you want to address my comments in separate PR then either open new issue linked to this or add tasks to current issue that allows you and me to track what is left to do.

LMK and I can approve this. Though likely want to have @akihikokuroda approval as well since some of your changes are in the `step.py` which I think he created and updated the most.

Thx

akihikokuroda (approved):

george-lhj:
added a separate issue #242 to address this

PR #237, opened 2025-02-18
patch method to handler user input

### LINKED ISSUE #238: sequence method failing due to new schema
We don't require an agent in a step which we did previously, so we must update the sequence method in Workflow to reflect and handle this case.

george-lhj:
fixes #238 
now that agent isn't a required param, the sequential method needed to be updated to reflect that and handle the case of user inputs which we did not have before.

### DIFF
--- a/maestro/src/workflow.py
+++ b/maestro/src/workflow.py
@@ -90,21 +90,29 @@ class Workflow:
                 return steps.index(step)
     
     def _sequence(self):
-        prompt = self.workflow["spec"]["template"].get("prompt", "")
-        steps = self.workflow["spec"]["template"].get("steps", [])
-        if not steps:
-            raise ValueError("Workflow is missing required 'steps' key in 'spec'")
-        step_results = {}
-        for i, step in enumerate(steps):
-            agent_name = step["agent"]
-            agent_instance = self.agents.get(agent_name)
-            if not agent_instance:
-                raise ValueError(f"Agent {agent_name} not found for step {step['name']}")
-            step_results[f"step_{i}"] = prompt
-            response = agent_instance.run(prompt)
-            prompt = response if isinstance(response, str) else response.get("prompt", prompt)
-        step_results["final_prompt"] = prompt
-        return step_results
+            prompt = self.workflow["spec"]["template"].get("prompt", "")
+            steps = self.workflow["spec"]["template"].get("steps", [])
+            if not steps:
+                raise ValueError("Workflow is missing required 'steps' key in 'spec'")
+            step_results = {}
+            for i, step in enumerate(steps):
+                agent_name = step.get("agent")
+                input_step = step.get("input")
+                if agent_name:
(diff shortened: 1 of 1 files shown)


maximilien (commented):
LGTM but let me let @akihikokuroda give you final OK

Also let's get in the habit of adding a test or expanding a test with every new feature. Does not need to be complicated, I would just add a simple 2 agent workflow with input prompt and validate and run it.

akihikokuroda:
Oh, I thought the sequence is just executing the agents in `agents` from the top to the bottom.   There is little differences between the sequence and the condition.  Can we eliminate the `strategy` or make the sequence simple not to use the `steps`?  @maximilien WDYT?

maximilien:
@akihikokuroda please show examples here to be clear what alternatives you are suggesting. Ideally criteria should be:

1. simplification (usually smaller)
2. avoiding repeating information
3. clarity in resulting workflow YAML

Show alternatives in different comment so we can discuss and / or vote. Thx

akihikokuroda:
For sequence strategy:
```
apiVersion: maestro/v1
kind: Workflow
metadata:
  name: maestro-deployment
  labels:
    app: example2
spec:
  strategy:
    type: sequence
  template:
    metadata:
      name: maestro-deployment
      labels:
        app: example
        use-case: test
    agents:
        - test1
        - test2
        - test3
    prompt: Tell me a joke about IBM
```
It executes test1, test2 and test3 agents in sequence.

akihikokuroda:
For condition strategy:
```
apiVersion: maestro/v1
kind: Workflow
metadata:
  name: maestro-deployment
  labels:
    app: example2
spec:
  strategy:
    type: condition
  template:
    metadata:
      name: maestro-deployment
      labels:
        app: example
        use-case: test
    agents:
      - agent1
      - agent2
      - agent3
    prompt: This is a prompt
    start: step1
    exception: step3
    steps:
      - name: step1
        agent: agent1
        condition:
          - if: (in.some_condition == True)
            then: step2
            else: step3
      - name: step2
	agent: agent2
      - name: step3
	agent: agent3
	condition:
          - case: (in.some_condition == True)
            do: step2
          - case: (in.some_condition == False)
            do: step3
          - default: step3
```

The `step` can have `name`, `agent`, `condition` and `input`.  The `name` is only required.

maximilien:
First, your YAML has issues (second example). Use ```yaml when adding which tells GH to pretty print and validate YAML or ```json or ```python etc

I would prefer not having to specify strategy at all. Could we manage that?

akihikokuroda:
> I would prefer not having to specify strategy at all. Could we manage that?

Yes, we can take out `strategy` from yaml schema and treat all workflow in the conditional code.

akihikokuroda:
@george-lhj Would you set the `condition` strategy for now and keep working on with demos?  I can take out `strategy` from schema and yamls and `_sequence` from workflow.py.

george-lhj:
yes, sounds good. closing

Issue #236, opened 2025-02-18
Add ability to loop in workflow


maximilien:
This came from discussion in: https://github.com/i-am-bee/beeai-labs/pull/234#issuecomment-2666884610

What about having a loop feature instead so something like:

```yaml
...
    steps:
      - name: step1
        agent: test1
      while: (len(input.list) == 0)
        do: step2
      - name: step2
        agent: test2
...

```

Of course we could also do it like this:

```yaml
...
    steps:
      - name: step1
         agent: test1
      loop: step2
        until: (len(input.list) == 0)
      - name: step2
         agent: test2
...
```

- [x] schema changes for `loop`
- [x] agree on approach with @akihikokuroda 
- [x] implement new loop structure in `workflow.py`
- [ ] add a specific test running `loop_workflow.yaml`

maximilien:
OK some changes to proposal but should work well. Proposing:

```yaml
...
    steps:
      - name: step1
        agent: test1
        loop: 
            step: step2
            until: (input.happy == True)
      - name: step2
        agent: test2        
...
```

Which leads to schema addition:

```json
                  "loop": {
                    "type": "object",
                    "description": "looping construct",
                    "items": {
                      "type": "object",
                      "properties": {
                        "step": {
                          "type": "string"
                        },
                        "until": {
                          "type": "string"
                        }
                      }
                    }
                  }
```

Let me push a PR with schema changes and example and work on changes to `workflow.py` to support once we all agree.

maximilien:
Reopen since there are remaining sub task not solved by PR #239

akihikokuroda:
#270 has a test for loop.

maximilien:
I believe this is now done. Closing. Thx @akihikokuroda

PR #239, opened 2025-02-18
First proposal for loops in workflow

maximilien:
* added changes to `workflow_schema.json`
* added test example `loop_workflow.yaml`
* verified that example passes validation
* corrected `workflow_schema.json` `$id` property to point to correct file
* small changes to `conditional_workflow.yaml` name
* 
When completed this will complete some tasks in issue #236 

CC: @akihikokuroda @george-lhj


maestro/tests/yamls/workflows/loop_workflow.yaml
```diff
@@ -0,0 +1,27 @@
...
+        use-case: test
+    agents:
+        - test1
+        - test2
+    prompt: Tell me a joke
+    steps:
+      - name: step1
+        agent: test1
+        loop: 
+            step: step2
```
akihikokuroda, line 24: I'm not sure if we want `step` here.  The `step` may have `loop`, `condition`, and `input` in addition to `agent`.  It may be complicated.  Can we change it to `agent`?

maximilien: OK let's ask @george-lhj for his opinion especially if he can change his demo YAML and show us.

It's an easy change on my side. George WDYT?

maestro/schemas/workflow_schema.json
```diff
@@ -92,7 +92,22 @@
...
+                      "type": "object",
+                      "properties": {
+                        "step": {
+                          "type": "string"
+                        },
+                        "until": {
+                          "type": "string"
+                        }
                       }
                     }
```
akihikokuroda, line 112: The `items` is not necessary here.
so it should be
```yaml
                    "properties": {
                      "step": {
                        "type": "string"
                      },
                      "until": {
                        "type": "string"
                      }
                    }
```
instead of
```yaml
                    "items": {
                      "type": "object",
                      "properties": {
                        "step": {
                          "type": "string"
                        },
                        "until": {
                          "type": "string"
                        }
                      }
                    }
```

maestro/tests/yamls/workflows/loop_workflow.yaml
george-lhj: Ya I think agent wouuld be good inside the loop, so it just executes the agent n times and doesn't need to worry about anything else

maestro/schemas/workflow_schema.json
maximilien: OK fair. Good feedack. Addressing

maximilien:
@akihikokuroda let's do a quick review and merge for this as I created sub-tasks in issue #236 so we can have different PRs for issue (by you or myself). Thx

akihikokuroda (approved):
LGTM

Issue #116, opened 2025-01-16
Update existing weather checker example for new CLI


psschwei:
xref: https://github.com/i-am-bee/bee-hive/tree/3d7d4d843128c76170b1707d017252ab10c5278d/examples/bee-hive/weather-checker

AngeloDanducci:
I think this should move to POC2 so we can refactor all of the demos to the new spec + cli inclusion.

Issue #242, opened 2025-02-19
add tests for summary demo


george-lhj:
add a run.sh and test.sh to run and test workflow using CLI. For test use --dry-run so no mock agents are used.

@maximilien could you explain more on this? I added a run and test that runs the mock workflow already using the cli for summary demo

george-lhj:
<img width="163" alt="Image" src="https://github.com/user-attachments/assets/0b6a1d46-5d6a-4435-a588-5946e9a84ed9" />
Inside summary ai, I've added run.sh and test.sh which you can either run or dry run by using `./demos/workflows/summary.ai/test_yaml/test.sh` (if we are in the maestro level)

maximilien:
Run them as part of CI as I mentioned in last PR and issue

maximilien:
Also since we want all demos to be somewhat uniform move `*.sh` files into root of the demo. Follow `weather-checker.ai` example in PR #243

PR #243, opened 2025-02-19
Runs the `weather-checker.ai` demo in CI

maximilien:
* does not call `run.sh` but `test.sh` instead which runs workflow as `--dry-run`
* `doctor.sh` only checks `maestro` CLI is there but more could do more in future
* need to call `run.sh` in future

fixes issue #116

### DIFF
--- a/.github/workflows/maestro_demo-tests.yaml
+++ b/.github/workflows/maestro_demo-tests.yaml
@@ -33,11 +33,10 @@ jobs:
         with:
           # Small model - 7GB max in this runner
           model: llama3.1:latest
-      # TODO: generalise for other demos 
-      - name: Run CrewAI workflow
+      - name: Run weather.ai workflow
         run: |
           cd maestro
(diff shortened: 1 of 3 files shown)


maximilien:
Can I get a check review on this @george-lhj

We need to do the same for the other demos. Thx

.github/workflows/maestro_demo-tests.yaml
george-lhj, line 1: By default shouldn't we set the maestro demo ollama model to 3.1? Most of the demos are using that model, so changing to 3.2 might change some output

george-lhj (approved):

george-lhj:
for #242, #244 I will just follow the same format here

maestro/demos/workflows/weather-checker.ai/test.sh
george-lhj, line 1: workflow misspelled in the output check

PR #248, opened 2025-02-20
Add integration tests for all demos

### LINKED ISSUE #244: Run all demos in CI
Run all the demos in CI as done for `weather-check.ai` in PR #243 

Likely do this adding a `tools/run-demos.sh` which looks for all `demos/workflows/*` and calling `doctot.sh` and `test.sh` for each. Then call the `tools/run-demos.sh` in the Demos GH action: `.github/workflows/maestro_demo-tests.yaml`

george-lhj:
Added doctor.sh, run.sh, and test.sh files for all demos. 
created tools/run-demos.sh to automatically run all the demos and improved the CI pipeline by ensuring maestro runs correctly in GitHub Actions.
fixes #244


maximilien (approved):
Love it. Only feedback (as a refactor) is to create `demos/tools` directory with common scripts and other "tool" and change the `run.sh` to import the right tools and pass their directory since that seems to be the main difference. Of course, if a script, e.g., `doctor.sh` is the same exactly then just call it from common.

PR #250, opened 2025-02-20
Add deploy class

akihikokuroda:
Add new `Deploy` class.

The constructor takes agent yaml, workflow yaml files and necessary runtime environment variables.
There are 2 deploy time environment variables. 

1. CONTAINER_CMD: command to build container image (docker, podman, nerdctl, ...) 
2. BUILD_FLAGS: flags necessary to build the container image (--namespace=k8s.io, ...

Here is an example
```
deploy = Deploy("../tests/examples/condition_agents.yaml", "../tests/examples/condition_workflow.yaml", "BEE_API_KEY=sk-proj-testkey BEE_API=http://192.168.86.45:4000")
```
It has 2 instance methods.

```
deploy.deploy_to_docker()
deploy_to_kubernetes()
```

After these method are executed, web interface is available at http/ <host address:port>/static/maestro.html
The host address and port are
Docker: 127.0.0.1:5000
Kubernetes: 127.0.0.1:"nodeport number of maestro service"

### DIFF
--- a/maestro/deployments/Dockerfile
+++ b/maestro/deployments/Dockerfile
@@ -43,12 +43,12 @@ RUN python3.11 -m ensurepip --upgrade
 RUN python3.11 -m pip install openai pyyaml python-dotenv flask --no-cache-dir --upgrade && \
     python3.11 -m pip uninstall -y pip
 
-COPY maestro ./maestro
-COPY entrypoint.sh .
-COPY run_workflow.py .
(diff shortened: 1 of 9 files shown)


maximilien (approved):
Love it. UI could be nicer but we can work on this after. Let's merge so I can integrate into CLI. Don't forget some tests... maybe integration test that deploys and checks URL and then shutsdown?

Issue #251, opened 2025-02-20
Create common folder for demo integration tests


george-lhj:
Currently, we have a `doctor.sh`, `test.sh`, and `run.sh` inside of each demo, `run-demos.sh` (inside tools) uses in order to locate and test the integration.

The information within the these files itself is almost all the same (except summary.ai's nested folder), and the actual file paths of the agents and workflows themselves. As a result, we can refactor to create common files to use.

PR #256, opened 2025-02-21
Common tests

george-lhj:
Removed nested folder, updated gh actions, run to reflect the refactor
related to #251 
after this will create common doctor, and test files so we can remove them from inside all the demos


maximilien (approved):
Looks good but why are we keeping the old YAMLs? I assume temp and will be deleted. If different versions (functionality-wise) then let's start versioning demo. Happy to discuss.

Let's merge and we can discuss next step

george-lhj:
the old yamls still have some agents and code i want to preserve for when i further implement the new summary demo after this issue. If they are no use at that time, I will delete it

george-lhj:
lmk what u think @maximilien

PR #260, opened 2025-02-24
Add parallel

akihikokuroda:
Adding **parallel** in step.  The parallel has a list of agent.  The agents are executed in parallel and the step waits until the all agent complete.

Here is the sample workflow yaml file.
```yaml
kind: Workflow
metadata:
  name: input_workflow
  labels:
    app: input
spec:
  template:
    metadata:
      name: input_workflow
      labels:
        app: example
        use-case: test
    agents:
        - test1
        - test2
        - test3
        - test4
        - test5
    prompt: Select a number 1 through 6
    exception: step4
    steps:
      - name: list
        agent: test1
      - name: parallel
        parallel:
        - test2
        - test3
        - test4
      - name: result
        agent: test5

```

### DIFF
--- a/maestro/schemas/workflow_schema.json
+++ b/maestro/schemas/workflow_schema.json
@@ -113,6 +113,13 @@
                         }
                       }
                     }
+                  },
+                  "parallel": {
+                    "type": "array",
+                    "items": {
+                        "type": "string",
+                        "description": "agent"
+                    }
                   }
                 },
                 "required": [
--- a/maestro/src/agent.py
+++ b/maestro/src/agent.py
@@ -4,6 +4,7 @@
 from abc import abstractmethod
 import os
 import pickle
+import asyncio
 
 class Agent:
     """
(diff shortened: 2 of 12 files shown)


maximilien (approved):
Love this @akihikokuroda I have some comments but let's merge as these could be done in separate issue. the main suggestions are:

1. move all agents file in new directory: `src/agents/` The rationale is to have all agent code there. Eventually we could do the same for other logical groups. But at least we could start with agents. Especially the case as there are many many agent frameworks and we'd want to have support to many.
2. We need better tests. The code you have now should be moved to a `TestParrralel` class and add some `test_xyz(...)` methods
3. Add tests that could verify that the agents are running in parallel. This is hard but we could start by having agents that just print their names and reading that the names are not in order... This could be done by running workflow a few times and checking order is not sequential. Other strategies could be used as well... but not urgent for now.

PR #264, opened 2025-02-25
fix temp directory

akihikokuroda:
Move the temp directory for the deploy to $(tempfile.gettempdir())"/maestro".  My Mac had "TMPDIR" defined.  It was the issue for my case.  

This is from the doc:
```
The directory named by the TMPDIR environment variable.
The directory named by the TEMP environment variable.
The directory named by the TMP environment variable.
A platform-specific location:
On Windows, the directories C:\TEMP, C:\TMP, \TEMP, and \TMP, in that order.
On all other platforms, the directories /tmp, /var/tmp, and /usr/tmp, in that order.
As a last resort, the current working directory.
```


maximilien (approved):
Thanks. I think this would work

PR #269, opened 2025-02-26
created prompts for meta agent agents.

george-lhj:
Found prompts and created the agents that replicate the process to create agent.yaml and workflow.yaml (the prompt I think is a bit too specific to be general, its more for the weather demo). I'm thinking might have to create another agent, or allow some type of user input? Unsure yet.

- [x] Next step is to create some type of workflow to execute these agents.

CC: @akihikokuroda @psschwei if you wanted to see current status; I am able to create the agents and workflow file using the meta agent, but the workflow file is too customized to just the weather case.

Current process outlined in the readme, after this is done need to try and generalize the workflow file somehow


maximilien (approved):
Love it. Let's generalize and move specific meta agents to demos. I would assume we will have a series until we find one that works more or less best for our demos. Good progress George. Leaving it open for others to take a look otherwise feel free to merge and move to next step.

PR #271, opened 2025-02-26
add prompt passing feature

### LINKED ISSUE #272: Add more customizable prompting in input function (in step)
The input function currently allows us to "stop" during the workflow and get a human input by specifying in the workflow: 

input: .....
{response}

Taking advantage of this 'input agent', we can add another feature with a different keyword that allows us to modify the prompts in the workflow so that we don't have to create connector agents, which do not give us the correct prompt most of the time. Instead, we can use a keyword like {CONNECTOR} to specify this particular action. Will greatly help the demos as it can give more customizable control during the workflow

george-lhj:
fixes #272 
Currently in the demo, the input keyword basically allows us to kind of "modify the prompt" a bit. In response to this, I've added an additional small feature, which is to pass a specific prompt using the {CONNECTOR} keyword to take advantage of this structure. 

Example usage in the meta agent demo:

agent1: creates the agent yamls necessary for a topic (in this case weather)
input: {CONNECTOR}.  /// because we don't want these files as output, but rather the prompt for the next agent2
agent: create the workflow yaml using what agents are being defined

### DIFF
--- a/maestro/src/step.py
+++ b/maestro/src/step.py
@@ -65,9 +65,10 @@ class Step:
         return default
 
     def input(self, prompt):
-        user_prompt = self.step_input.get("prompt").replace("{prompt}", str(prompt)) 
-        response = input(user_prompt) 
+        user_prompt = self.step_input.get("prompt").replace("{prompt}", str(prompt))
         template = self.step_input.get("template")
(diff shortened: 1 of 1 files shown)


akihikokuroda (approved):
LGTM

PR #282, opened 2025-03-03
Generalizing agent creation within meta-agents

### LINKED ISSUE #280: Generalize agents creation in meta agents
Currently, meta agent is able to properly create agents, but in the workflow, we are specifically creating the agents for demo. We need to generalize this so that all demos can use this.

To do this, we probably can introduce a prompt_formatting input agent, which the user will describe how many agents and what each agent does. The output should be the desired prompt of the current inital agent.

george-lhj:
Previously was fixed for weather demo case, but now added another agent which would take in the user prompt for any case, and format it correctly no matter which demo.

Fixes #280, still need to do this for the workflow creation.


maximilien (approved):
Love it. Need to test of course ;)

PR #285, opened 2025-03-05
fix default of env argument for deploy

akihikokuroda:
fix default of env argument for deploy

### DIFF
--- a/maestro/src/deploy.py
+++ b/maestro/src/deploy.py
@@ -63,7 +63,7 @@ def update_yaml(yaml_file, str_envs):
         yaml.safe_dump(data, f)
 
 class Deploy:
-    def __init__(self, agent_defs, workflow_defs, env=None, target=None):
+    def __init__(self, agent_defs, workflow_defs, env="", target=None):
         self.agent = agent_defs
         self.workflow = workflow_defs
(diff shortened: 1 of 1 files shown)


maestro/src/deploy.py
```diff
@@ -63,7 +63,7 @@ def update_yaml(yaml_file, str_envs):
         yaml.safe_dump(data, f)
 
 class Deploy:
-    def __init__(self, agent_defs, workflow_defs, env=None, target=None):
+    def __init__(self, agent_defs, workflow_defs, env="", target=None):
```
maximilien, line 66: Shouldn't this be `env={}` instead?

maximilien (changes requested):

maestro/src/deploy.py
akihikokuroda: No, it's expecting a string.  It is like a command string "key1=value1 key2=value2".

maximilien: Ok sounds good. This could be my error since I expected it to be a dictionary. Thx

maximilien:
LGTM

maximilien (approved):
LGTM

Issue #281, opened 2025-03-03
Generalize workflow creation in meta agents


george-lhj:
Currently, meta agent is not able to create a workflow generally, since in the prompts we are specifically using weather demos for the workflow.

To improve this, we probably can introduce another agent or use the prompt_formatting input agent that was previously created, which the user will describe the workflow. The output should be the desired workflow.yaml file of the demo

george-lhj:
PR #286  will give the generalization ability, but create a new issue of splitting the workflow.yaml into 2

PR #286, opened 2025-03-05
create generalization for workflow, agents

george-lhj:
This is solving issue #281 
 
If you look at the workflow, I've separated the original workflow into 2, `workflow.agent` which creates the agents.yaml file and also `workflow.workflow` which creates the workflow.yaml file. This is becuase it gives better control over the prompt as well as simplifies the execution.

I've created the original prompt so that both of these actually have the same starting prompt, which means they can be either:
(1) run in parallel to be executed from the original prompt (so that both these agents use that og prompt, and not the output of whatever gets executed first)
(2) create another overarching workflow that calls on both of these workflows, so we need to add support for a workflow executing another workflow? As I would have an overarching workflow.yaml that has one prompt that should be passed and execute the 2 current workflow files.
(3) other ideas

I'm not too sure which one to go for, seems like the parallel right now may be a bit slow? So I'm leaning toward the 2nd idea. WDYT? And is this possibly creating any new features in the backend we need to address?


george-lhj:
Note: test currently failing because I have renamed the workflow files, easy fix once a solution is decided.

psschwei:
I'm having a bit of trouble following what the different options you laid out entail (possibly just due to me not being familiar with this workstream). Could you maybe spell them out in a bit more detail?

george-lhj:
> I'm having a bit of trouble following what the different options you laid out entail (possibly just due to me not being familiar with this workstream). Could you maybe spell them out in a bit more detail?

Sure sorry.

Right now, to execute a demo, we would simply run the command `maestro run agent.yaml workflow.yaml`. Then, this creates the agents based on the agent yaml file and then executes the workflow based on the workflow.yaml file. 

Now, we have a new goal which is to have an meta-agent that given a prompt, we can create the agents/workflow yamls automatically. 

I am able to generalize the process for the creation of agents currently, but its hard to generalize for workflow if its executed after the creation of agents (as sequentially or condionally, the output of the previous step should be the agents.yaml file, which doesn't make sense for the prompt of creating the workflow.yaml file.

Instead, I have split these processes to use the same prompt, but in 2 different workflow files because their process is different. Because they have the same prompt, I think we can execute them in parallel, or instead create an overarching workflow that would execute both of these using the same prompt. 

Is this a bit more clear? Sorry, the nature of meta-agent is just a bit recursive so its a bit confusing at first

psschwei:
Thanks, that's helpful.

I wonder if the agent creation step should have two outputs: 
1) the agent yaml, and
2) a list of agent names

They'd get returned as a dictionary, and then we could pass (2) as part of the input for the workflow creation step.

Alternatively, maybe there's a need for a third agent that can scan the agent file and return the name(s) of available agents. 

It is also very possible I'm drastically overthinking things and there's a simpler solution. Just wanted to call that one out too :)

george-lhj:
> Thanks, that's helpful.
> 
> I wonder if the agent creation step should have two outputs:
> 
> 1. the agent yaml, and
> 2. a list of agent names
> 
> They'd get returned as a dictionary, and then we could pass (2) as part of the input for the workflow creation step.
> 
> Alternatively, maybe there's a need for a third agent that can scan the agent file and return the name(s) of available agents.
> 
> It is also very possible I'm drastically overthinking things and there's a simpler solution. Just wanted to call that one out too :)

This makes sense, but currently based on how we have defined our process in workflow.py, step.py, etc, we directly pass the output of the previous agent into the next agent as the prompt, which means we would have to pass the entire dictionary object as the prompt which isn't necessary and adds complexity. Of course, we can add more changes inside the backend, but thats prob a question for @maximilien or @akihikokuroda

maximilien (commented):
Love this... I assume you have more work since it's DRAFT. I'll keep an eye and review soon as you update. Really excited to try this. Nice progress @george-lhj

george-lhj:
@maximilien

maximilien:
@george-lhj one ideal flow would be:

1. call meta agent workflow passing the description of the workflow needed, this is a text of file with the text
2. meta agent uses to create suggested agents.yaml and workflow.yaml
3. meta agent ask user any clarifying questions needed
4. using answers in 3 then meta agent refines the agents.yaml and workflow.yaml in 1 to generate new versions
5. meta agent repeats 2-4 as needed
6. once no more questions need to be answered meta agent tells user how to run the workflow
7. user tries to run workflow and repeats 1-7 if needed

WDYT?

CC: @psschwei @akihikokuroda

maestro/src/agents/meta_agent/agents.yaml
```diff
@@ -162,4 +162,95 @@ spec:
...
+                    Each step must:
+                    Reference a valid agent from agents.yaml.
+                    Have a name that describes its function.
+                    The final output should be a valid structured workflow in YAML format, please make it easily readable in a code block.
+                    ```
+
+                    **Example User Input:**
+                    ```
+                    number of agents: 2
+                    agent1: weather_fetcher – Retrieves weather data for a given location using OpenMeteo.
```
maximilien, line 204: should we say "OpenMeteo tool" ?

maximilien (dismissed):
Love the start. Of course in my mind the since we want to start with a textual description of the workflow. Perhaps a "pre-process" agent is needed that converts the workflow English description into a structured prompt as you have here. Then the structured prompt could then be used to start generating the YAMLs and and iterate as I suggested. WDYT?

maestro/src/agents/meta_agent/agents.yaml
george-lhj: ya

george-lhj:
> Love the start. Of course in my mind the since we want to start with a textual description of the workflow. Perhaps a "pre-process" agent is needed that converts the workflow English description into a structured prompt as you have here. Then the structured prompt could then be used to start generating the YAMLs and and iterate as I suggested. WDYT?

Sure I agree, but for now I want to replicate the other demos. This version makes it less complicated so I can easily test and control the prompts. I think after that it would be reasonable to implement this.

My suggested steps are:

- [ ] merge this generalized prompts/workflow
- [ ] decide how we want to implement the combination of multiple workflows (parallel, another workflow that calls these 2, etc)
- [ ] refactor/restructure so that only the general files are in this folder, and create another folder in demos for meta agent demos
- [ ] replicate certain demos
- [ ] go back, edit agents so that we can have a textual description

maximilien (approved):
LGTM to move fast

PR #287, opened 2025-03-05
Shorten parallel test

### LINKED ISSUE #284: Can we make parallel tests faster?
Currently running the new parallel tests take almost a minute. In particular the one outputing the following:

```bash
Mock agent:Loading code_interpreter
Mock agent:Loading test
Mock agent: name=test5, model=meta-llama/llama-3-1-70b-instruct, description=this is a test, tools=[], instructions=print(input)
🐝 Running test1...
5
🐝 Response from test1: Select a number 1 through 6
🐝 Running test2...
🐝 Response from test2: test2
🐝 Running test3...
🐝 Response from test3: test3
🐝 Running test4...
🐝 Response from test4: test4
🐝 Running test5...
['test2', 'test3', 'test4']
🐝 Response from test5: ['test2', 'test3', 'test4']
```

Ideally all tests (especially unit and some integration tests) should be as fast as possible to encourage devs to run them often. Since we have few tests and as we add more the test suite will become slower and slower we should try to make tests fast. If we cannot, and they must be slow, then let's figure out how to move them to CI.

CC: @george-lhj

akihikokuroda:
fix #284

### DIFF
--- a/maestro/tests/yamls/agents/dry_run_paralle_agent.yaml
+++ b/maestro/tests/yamls/agents/dry_run_paralle_agent.yaml
@@ -30,7 +30,7 @@ spec:
     - test
   instructions: |
     import time
-    time.sleep(15)
+    time.sleep(5)
     input = "test2" 
 
 ---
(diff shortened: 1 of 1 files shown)


maximilien (approved):
Love it... we might even be able to get it shorter with 1 or 2 seconds instead. But let's try with these. Thx

PR #294, opened 2025-03-08
Add agent

george-lhj:


maximilien (approved):
Love it @george-lhj

Definite step in the right direction and I can see this trail of meta-agents in the end to be super useful as we keep improving.

Let's merge!

Issue #291, opened 2025-03-06
pre demo completion tasks


george-lhj:
- [x] fix gh actions to allow for multiple yamls checking in a demo
- [x] merge this generalized prompts/workflow
- [x] go back, edit agents so that we can have an english description that generates the current structured prompt used for generating the yaml files
- [x] refactor/restructure so that only the general files are in this folder, and create another folder in demos for meta agent demos (v1, and for each new version we will create a different file)
- [ ]  allow for user to check and fix prompt before entering the workflow.

maximilien:
Can we close this @george-lhj  since 3/12 demo is done?

PR #309, opened 2025-03-13
prompt adjustments 

george-lhj:
fixe #291 

For the demo I had a few prompts that were working better so edited them.

I am wondering should I add the generated agents.yaml/workflow.yaml files that meta agent creates? Or keep it out


maximilien (approved):
LGTM

Issue #297, opened 2025-03-11
Complete Workflow::to_mermaid markdown with other workflow elements


maximilien:
PR #296 solved issue #290 but needs to be enhance to include markdown representation for:

- [x] add CI to run `maestro mermaid <workflow.yaml>` on all workflow YAMLs
- [x] workflow `condition` (case/do/default and if/then/else)
- [x] workflow `exception` 
- [x] workflow `on` events sequence diagram 
- [x] workflow `parallel` sequence diagram 
- [x] workflow `loop` sequence diagram 

- [x] workflow `exception` flowchart
- [x] workflow `on` events flowchart
- [x] workflow `parallel` flowchart
- [x] workflow `loop` flowchart

CC: @akihikokuroda

psschwei:
Once we can represent all workflows as mermaid markdown, I think it would be worth asking the question of can we go the other way? Submit a workflow to maestro as mermaid markdown and have it parse it into the correct order...

psschwei:
https://github.com/i-am-bee/beeai-labs/issues/303

maximilien:
```yaml
apiVersion: maestro/v1
kind: Workflow
metadata:
  name: maestro-deployment
  labels:
    app: example2
spec:
  template:
    metadata:
      name: maestro-deployment
      labels:
        app: example
        use-case: test
    agents:
      - agent1
      - agent2
      - agent3
    prompt: This is a prompt
    start: step1
    exception: step3
    steps:
      - name: step1
        agent: agent1
        condition:
          - if: (in.some_condition == True)
            then: step2
            else: step3
      - name: step2
        agent: agent2
      - name: step3
        agent: agent3
        condition:
          - case: (in.some_condition == True)
            do: step2
          - case: (in.some_condition == False)
            do: step3
          - default: step3
```

generates this piece of mermaid sequenceDiagram

```mermaid
sequenceDiagram
participant agent1
participant agent2
participant agent3
agent1->>agent2: step1
agent2->>agent3: step2
agent3->>agent3: step3
```

and this piece of mermaid flowchart

```mermaid
flowchart LR
agent1-- step1 -->agent2
agent2-- step2 -->agent3
agent3-- step3 -->agent3
```

maximilien:
This is complete for sequence diagram but some flowcharts missing.

PR #316, opened 2025-03-14
feature: add support for condition (case/do/default and if/then/else)

maximilien:
Splitting support for `condition` in mermaid into:

- [x] `case/do/default` sequenceDiagram and flowchart
- [x] `if/then/else` sequenceDiagram
- [x] `if/then/else` flowchart

When done this will complete `condition` subtask in issue #297

### DIFF
--- a/maestro/src/mermaid.py
+++ b/maestro/src/mermaid.py
@@ -71,12 +71,14 @@ class Mermaid:
             if step.get('condition'):
                 for condition in step['condition']:
                     condition_expr = ''
+                    # generate the case / do / default
                     if condition.get('case'):
                         condition_expr, do_expr = condition['case'], condition['do']
                         if condition.get('default'):
                             condition_expr = 'default'
                             do_expr = condition['default']
                         sb += f"{agentL}->>{agentR}: {do_expr} {condition_expr}\n"
+                    # generate the if / then / else
                     elif condition.get('if'):
                         if_expr, then_expr, else_expr = condition['if'], condition['then'], ''
                         if condition.get('else'):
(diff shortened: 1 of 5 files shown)


maximilien:
@psschwei or @george-lhj this is ready for review

PS: @akihikokuroda on vacation so one of you need to review and give me OK or feedback. Thx

maximilien:
When merged this solves the:

- [x] workflow condition (case/do/default and if/then/else)

subtask of issue #297

maximilien:
Share an example workflow please not just hypothetical. Thx

maestro/src/mermaid.py
```diff
@@ -60,12 +60,34 @@ def __to_sequenceDiagram(self, sb):
...
+                        if_expr, then_expr, else_expr = condition['if'], condition['then'], ''
+                        if condition.get('else'):
+                            else_expr = condition['else']
+                        sb += f"{agentL}->>{agentR}: {if_expr}\n"
+                        sb += f"alt if True\n"
+                        sb += f"  {agentL}->>{agentR}: {then_expr}\n"
+                        if condition.get('else'):
+                            sb += f"else is False\n"
+                            sb += f"  {agentR}->>{agentL}: {else_expr}\n"
+                        sb += f"end\n"
```
psschwei, line 90: I will try to review later, but would it be possible to add comments for this block? (maybe an llm could do it?) I am not able to easily follow the logic of what this code is doing.

maestro/src/mermaid.py
```diff
@@ -90,7 +112,24 @@ def __to_flowchart(self, sb, orientation):
                 sb += f"{agentL}-- {step['name']} -->{agentR}\n"
             else:
                 sb += f"{agentL}-- {step['name']} -->{agentL}\n"
-            i = i + 1        
+            # if step has condition then add additional links
```
psschwei, line 115: also a comment here would help

maestro/src/mermaid.py
maximilien: Happy to add but the code speaks for itself really ;)

maximilien: Added some comments in code but to really understand look at the tests for conditional in `tests/test_mermaid.py` since this includes example output for some of the test workflows with conditional. Otherwise, rest of code pretty straightforward

maximilien:
@george-lhj and @psschwei i need this moving faster. I have code to complete issue #297 and if we merge out of order it will be a mess. So please open separate issues or let me know what you want change. Thx

psschwei (approved):

maximilien:
Thanks @psschwei. Merging.

@george-lhj please open a separate issue with your comment and make sure to add a sample workflow with generated mermaid and what is wrong and what you would expect. I'd be happy to address it ASAP. Thx.

PR #323, opened 2025-03-19
Fix activity demo 

akihikokuroda:
There were some issues in the activity-planner.ai demo agent and workflow yaml files.


maximilien (approved):
LGTM

PR #334, opened 2025-03-21
typo in file name causing error

george-lhj:
realized the agents file has ".." instead of just one, causing an error in deployment


psschwei (approved):

maximilien (commented):
LGTM

PR #342, opened 2025-03-24
fixed newline error

### LINKED ISSUE #339: bug: add new line between agent executions
Currently when running deployed agents, the streaming output mangles the different running agents.
Add a `\n` after agent output.

george-lhj:
closes #339 

Previous code is replacing \n with \ndata: (which is markdown line break + next data prefix), but it's putting it all in one yield, which causes multiple lines to be appended as one message


maximilien (approved):
LGTM

PR #345, opened 2025-03-24
option for default prompt

### LINKED ISSUE #332: deployment runs by default
Even without putting a prompt, the deployment is running the NY prompt, which means by default its getting the prompt from the yaml without waiting for the prompt from the user

In the image below, I put in a prompt for san jose, which is why we can see that we are running 2 instances of the weather demo at the same time with different prompts

Image

george-lhj:
fixes #332 
added new --auto-prompt flag to the maestro deploy command. When present, the deployed Maestro UI will automatically start the workflow using the prompt specified in the YAML file. 

By default upon load, we will need to wait for the user to enter  the prompt if the flag is not added

Added --auto-prompt flag to maestro deploy CLI definition (maestro.py, commands.py)
Set AUTO_RUN=true as an environment variable if --auto-prompt is used
Updated api.py to conditionally run the workflow on page load if AUTO_RUN is set or a prompt is manually submitted

### DIFF
--- a/maestro/cli/commands.py
+++ b/maestro/cli/commands.py
@@ -224,12 +224,12 @@ class DeployCmd(Command):
     def __deploy_agents_workflow(self, agents_yaml, workflow_yaml, env):
         try:
             if self.docker():
-                deploy = Deploy(agents_yaml, workflow_yaml, env)
+                deploy = Deploy(agents_yaml, workflow_yaml, env, autoprompt=self.auto_prompt())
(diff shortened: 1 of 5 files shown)


maximilien (changes requested):
Please add a test @george-lhj and we will be good to go. Keep it simplete see my `test_commands.py` deploy test and add a new one for `--auto-prompt`. Thx

PR #351, opened 2025-03-25
Agent renaming

### LINKED ISSUE #344: Renaming steps, agents in meta agents, weather
Better prepare for demo so that step names make more sense (easier to follow in the mermaid charts)
and the agent names describe the job they are doing

george-lhj:
in preparation for demo, have renamed agents in weather demo, meta agents, and meta agents output so they reflect their usage better. Also, this is reflected in the creation of the mermaid graphs, as seen here:
Screenshot 2025-03-25 at 11 40 20 AM

closes #344


maximilien (approved):
Thx

LGTM

PR #360, opened 2025-03-26
added list of FRAMEWORKS and EMOJIS

maximilien:
Remaining TODOs:

- [ ] switch code to use `FRAMEWORKS`
- [ ] add common `Agent::print` method that prepends the correct `EMOJI` to message to be printed depending on actual agent framework
- [ ] add tests


george-lhj (approved):

psschwei:
@maximilien looks like you need to sign your commits

image

(btw, the bot was re-added automatically, so must be a LF thing)

PR #361, opened 2025-03-26
prompt updates

george-lhj:
update activity planner to get more consistent results via CLI in preparation for demo


maximilien (approved):
LGTM

PR #371, opened 2025-03-28
Bug: using json serialization instead of fstrings to prevent newline errors

### LINKED ISSUE #370: Maestro create not parsing '|' anymore
Previously, many of the demos/agent.yaml files are created like this, as we can see in our main branch:

Image

This | allows us to use newlines, making spacing a bit easier and simpler to define complex instructions for workflows.

recently, this | is failing to parse now, so the maestro create command does not work when this occurs

george-lhj:
fixes #370 

The agent creation fails with a KeyError for "id" because the multi-line instructions (using a plain vertical bar) retained an extra newline. This extra newline caused the manually constructed JSON payload to be malformed, resulting in an unexpected response from the API.

Switched from manually building the JSON string using an f-string to constructing a Python dictionary and serializing it with json.dumps. This ensures all special characters and newlines in fields like "instructions" are properly escaped.
Optionally, applied .strip() to the instructions value to remove any unwanted leading/trailing whitespace.

Also changed the top_p (which is the probability distribtion param) to be 0.8 instead of 1; since we want more deterministic answers

### DIFF
--- a/maestro/src/agents/bee_agent.py
+++ b/maestro/src/agents/bee_agent.py
@@ -4,7 +4,7 @@
 import os, dotenv
 import asyncio
 import requests
-
+import json
 from openai import AssistantEventHandler, OpenAI
 from openai.types.beta import AssistantStreamEvent
 from openai.types.beta.threads.runs import RunStep, RunStepDelta, ToolCall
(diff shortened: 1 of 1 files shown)


maximilien (approved):
Nice work. Can we make sure @akihikokuroda gives his OK before merging. He's back Monday so let's wait for him to merge.

My main reason to wait is I don't understand why this was not identified before? Was multiple lines never used for instructions? Or we never tested this path until now?

Anyhow nice work finding this.

george-lhj:
I have no clue why this wasn't found before, I had pretty complex instructions since the beginning of building meta agent with multiple lines spanning for the instructions (thats why in the main the instructions all have this). 

However, I was creating the agents again last night for the demo and it didnt work so must have been a recent change (but I have no idea when that caused this)

maximilien:
Ok. Let's let @akihikokuroda take a look on Monday and approve. Have great weekend all.

akihikokuroda (approved):
LGTM

PR #374, opened 2025-03-31
reuse existng agent

### LINKED ISSUE #362: bug: deployment creates new agents every run
Current everytime we are deploying, agents are automatically created with default tools (Openmateo, duckduckgo, and code intepreter) enabled. 

For the weather demo, this is working (not perfectly, because we don't need the duckduckgo) but we can get away with it.

However, for more complex demos, even starting from activity planner: we need specific tools that are not enabled by default. Therefore creating these agents when we deploy gives us no chance to enable it. As a workaround, we should check if in the localhost we already have an agent with the name created, and if there is, use that.

I am proposing to use this solution because for even. more complex demos, like meta agents, we need not only specific tools, but custom tools (ones with code). Right now we don't have the ability to auto create these tools and assign to agent at deploy. Therefore the easiest solution is to create them using the CLI, enable tools in the UI, and then check/use them using deploy so they aren't created

akihikokuroda:
fix #362 

This is for beeai agent only.  When the agent with the same name and model exists, the agent is used instead of creating a new one.

### DIFF
--- a/maestro/src/agents/bee_agent.py
+++ b/maestro/src/agents/bee_agent.py
@@ -28,6 +28,17 @@ class BeeAgent(Agent):
         super().__init__(agent)
     
         url = f'{os.getenv("BEE_API")}/v1/assistants'
+        headers = {
+            'accept': "application/json",
+            'Authorization': "Bearer sk-proj-testkey",
+            'Content-Type': "application/json"
+        }
+        response = requests.request("GET", url, headers=headers).json()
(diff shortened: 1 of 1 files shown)


psschwei (approved):

george-lhj:
Ive tested this on the UI, which now doesn't have to create new agent. Thanks!

george-lhj (approved):

PR #377, opened 2025-03-31
Exception support

### LINKED ISSUE #354: feature: implement workflow global exception
When workflow step end up in exception, the step / agent specified in the `exception` should be called to handle exception.

akihikokuroda:
fix #354 
The agent specified as the exception in the workflow is called when exception happens.


maestro/src/workflow.py
```diff
@@ -104,8 +105,17 @@ def to_mermaid(self, kind="sequenceDiagram", orientation="TD") -> str:
 
     async def run(self):
         """Execute workflow."""
-        self.create_or_restore_agents(self.agent_defs, self.workflow)
-        return await self._condition()
+        try:
+            self.create_or_restore_agents(self.agent_defs, self.workflow)
+            return await self._condition()
+        except Exception as err:
+            if self.workflow["spec"]["template"].get("exception"):
```
psschwei, line 112: do we need to explicitly handle the case where there is not exception in the template?

akihikokuroda: I'm not sure.  How about re-throwing the exception?

psschwei: I think that works :+1:

maximilien (changes requested):
Can you add a simple test @akihikokuroda? for instance force an exception in agent then catch and print and done. Or similar then we are good to go. Thank you.

maximilien:
@psschwei any chance you can check again on DCO? Why did this come back?

psschwei:
> @psschwei any chance you can check again on DCO? Why did this come back?

I don't know. I removed it a few minutes ago but seems it was automatically added back. So I guess there's some org-level policy that's overriding things.

psschwei:
I just removed it again, will see how long it lasts

Issue #365, opened 2025-03-28
feature: create a special meta-agents deployed UI


maximilien:
Currently the workflow deployed UX is designed for single workflow in development mode.

The meta-agents workflows are inherently meant to be iterative and also since they themselves create workflows there is a need for a different UX for meta-agents that helps the user execute the high-level meta-agents process. Summarized here:

1. Take English description of intended workflow with clear input and output
2. NLP meta-agent: description into English prompts to next meta-agents
3. Format / Create meta-agents: generate workflow.yaml and agents.yaml 
4. Iterate 2-3 to have a running workflow.yaml
5. Human user evaluates and retries 1-4 until “happy”

@george-lhj let's use this issue to come up with high-level UX and visuals and implement in streamlit.

CC: @akihikokuroda

maximilien:
this is now fixed

PR #380, opened 2025-03-31
feature: first version of meta-agents streamlit UI/UX

maximilien:
Still TODOs:

- [x] since meta-agents prompts are multiple lines need to change the default `st.chat_input()` to be multilines
- [x] no UI yet for the generated workflow but should be easy if we can capture when tab is selected and pass the files
- [ ] does not show mermaid for the meta-agents workflow workflow
- [ ] needs testing

@george-lhj this is not fully ready but worth trying at least to test the meta-agents workflow for generating the agents. You can try it with:

```bash
./maestro meta-agents tests/meta_agents/simple_prompt.txt
```

OR change ` tests/meta_agents/simple_prompt.txt` to a different file. If file does not exist it should default to a simple prompt.

When complete and merged this will close issue #365

### DIFF
--- a/maestro/cli/commands.py
+++ b/maestro/cli/commands.py
@@ -23,7 +23,7 @@ from jsonschema.exceptions import ValidationError, SchemaError
 
 from src.deploy import Deploy
 from src.workflow import Workflow, create_agents
-from cli.common import Console, parse_yaml
+from cli.common import Console, parse_yaml, read_file
 from cli.streamlit_deploy import deploy_agents_workflow_streamlit
 
 # Root CLI class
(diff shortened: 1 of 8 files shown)


maximilien:
@george-lhj added a text area with prompt so the input chat can be what you are asking agent to do

maximilien:
OK @george-lhj pushed a few more changes. See if you can test the agents creation workflow. If we are good there then we can complete the rest for sure since the code is almost the same.

Key think is to pass the prompt if user changes the text area. I did not fix that yet. But feel free to push changes to this PR.

maximilien:
@psschwei why are we getting damn DCO check again?

maximilien:
@george-lhj pull this new version. Should be able to run meta-agents to generate agents.yaml and workflow.yaml with initial prompt that you can edit. LMK if it works. 

I will work on last bit. Which will allow you to load the YAML files. As a first attempt.

george-lhj:
ok, the deploy gets agents.yaml and workflow.yaml dynamically, so I will finish updating the prompts into the src file. For now will approve.

george-lhj (approved):

maximilien:
opening "does not show mermaid for the meta-agents workflow workflow" as different issue

PR #388, opened 2025-04-02
Bug: workflow deployment not rendering

george-lhj:
agent definintion missed `---` so workflow agent file was not being read properly.

Changed the location of workflow yaml to the correct place


maximilien (approved):
LGTM

PR #389, opened 2025-04-02
Fix streaming output

### LINKED ISSUE #372: feature: allow streaming inside of streamlit
Currently, when we input a prompt, we don't see the output of all the agents/steps until the very end once its done running.

User likely wants to see output at each step, in case there is an error, or wants to see the output.

akihikokuroda:
fix #372 

This pr also have fix to make the prompt input work

### DIFF
--- a/maestro/cli/streamlit_deploy.py
+++ b/maestro/cli/streamlit_deploy.py
@@ -119,11 +119,17 @@ def deploy_agents_workflow_streamlit(agents_file, workflow_file):
 
             # stream response
             while True:
-                message = output.getvalue()
+                message = ""
+                lines = output.getvalue().splitlines()
+                for line in lines:
+                    message = message + f"{line}\n\n"
                 message_placeholder.markdown(message)
                 time.sleep(1)
                 if not thread.is_alive():
-                    message = output.getvalue()
+                    message = ""
+                    lines = output.getvalue().splitlines()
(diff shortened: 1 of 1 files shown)


maximilien (commented):
let me try this @akihikokuroda, thank you.

maximilien (changes requested):
Unsure if this worked @akihikokuroda seeing thee issues:

1. seems to be slower starting but could be my environment
2. the agent response needs new line (see screenshot)
3. I am process of refactoring the deploy streamlit code. See the file `cli/streamlit_workflow_ui.py` which contains the new refactorng code. You need to change that too so George can test meta-agents deploy which uses the refactored code.

@george-lhj worth for you to review and make sure it works for you once Aki addresses 3.

akihikokuroda:
I run the activity-planner.ai.  It formats fine.
Screenshot 2025-04-02 at 3 49 58 PM

akihikokuroda:
The DRY_RUN doesn't format well.

maximilien:
That must be it since Chrome same.

george-lhj:
The meta specific agents UI I believe has a separate deploy, so you have to change for that file as well. Not sure if that should be in this PR, or another

akihikokuroda:
@maximilien I prefer another PR for the meta agent UI.

Issue #349, opened 2025-03-25
enhancement: remove need for None in `maestro run workflow.yaml`


maximilien:
When running maestro workflow.yaml without `agents.yaml`

```bash
maestro run None workflow.yaml
```

There is no need for `None` since CLI could see that the file pass is a workflow and just run it. So user can do:

```bash
maestro run workflow.yaml
```

With same result

george-lhj:
this should be closed

PR #391, opened 2025-04-03
fixes issue #349

maximilien:
* [91mError: 'file_path'[0m or [91mError: 'file_path'[0m assuming agents used already created
* added tests for both cases
* renamed all CLI test to include  in name


george-lhj:
this is causing me issues, but im not sure what yet
`Error: Unable to run workflow: No module named 'src.agents.bee_agent'`

george-lhj (approved):
issue was that it doesn't recognize any of the agents that have already been created, so i need to recreate all the agents once this is merged

george-lhj:
so how come we don't update the `maestro deploy` to have the same logic if the run is being updated to this?

george-lhj:
finished recreating and running all the demos. works

PR #396, opened 2025-04-04
Add show diagram button to fix issue 386

### LINKED ISSUE #386: bug: meta agents workflow workflow does not show mermaid diagram
This was a subtask in PR #365

akihikokuroda:
fix #386 

The mermaid diagram must be drawn in a tab when the tab is displayed.  To make it, this PR add a "Show Dialog" button so that the tab is displayed when the diagram is drawn.

Please check if this is an acceptable solution.


george-lhj (commented):
nice, it this is fixed for all my tests. @maximilien wdyt? the tradeoff is that by default the diagram is not displayed

maximilien (approved):
Thanks for this Aki. @george-lhj can you also review and make sure this works on live demo. And when you have minute record again so I can update slides:)

Issue #394, opened 2025-04-04
bug: meta-agents should generate agents/workflow file


maximilien:
Meta agents should optionally create the `agents.yaml` and `workflow.yaml` files. As default just use the directory where maestro CLI is ran. That can be found in `sys.argv[0]`

george-lhj:
do you recommend creating an agent in order to take the output and try to download as a file? Or just write a function that does it?

maximilien:
I would add an agent with tool to write. Reason is we might want different agents to commit to GitHub for instance in future etc :)

PR #403, opened 2025-04-08
feature: load and create generated <agents/workflow>.yaml files

### LINKED ISSUE #395: feature: meta agents UI should auto load generated agents / workflow files
After issue #386 is solved the meta agents UI should auto load the generated `agents.yaml` and `workflow.yaml` files.

akihikokuroda:
Fix #395 

This PR adds "Load generated files" button in the "Generated" tab.  This expects the generated file names are 
```
    generated_agent = "generated_agent"
    generated_workflow = "generated_workflow"
```
in the working directory.

### DIFF
--- a/maestro/cli/streamlit_meta_agents_deploy.py
+++ b/maestro/cli/streamlit_meta_agents_deploy.py
@@ -11,6 +11,19 @@ from streamlit_workflow_ui import StreamlitWorkflowUI
 from cli.common import Console, parse_yaml, read_file
 
 def deploy_meta_agents_streamlit(prompt_text_file):
+    print("!!!!")
+    generated_agent = "generated_agent"
+    generated_workflow = "generated_workflow"
+
+    def load_generated():
+        print(os.getcwd())
+        if os.path.exists(generated_agent):
+            agent_file = generated_agent
+            st.session_state.agent_file = agent_file
+        if os.path.exists(generated_workflow):
+            workflow_file = generated_workflow
+            st.session_state.workflow_file = workflow_file
+
     prompt = "Enter your maestro meta-agents prompt here"
     try:
         prompt = read_file(prompt_text_file)
@@ -49,11 +62,18 @@ def deploy_meta_agents_streamlit(prompt_text_file):
 
     with generated_workflow_tab:
         st.header("Meta-agents generated ⌇ -> Workflow")
+        # Add load generated file button
+        st.button('Load generated files', on_click=load_generated)
         agent_file = st.file_uploader("Choose agents.yaml", key='agents.yaml')
         workflow_file = st.file_uploader("Choose agents.yaml", key='workflow.yaml')
 
         if agent_file is not None and workflow_file is not None:
             ma_gent_workflow_workflow_ui = StreamlitWorkflowUI(agent_file.getvalue().decode("utf-8"), workflow_file.getvalue().decode("utf-8"), '', 'Maestro meta-agents generated workflow')
+            ma_gent_workflow_workflow_ui.setup_ui()
+        elif "workflow_file" in st.session_state and "agent_file" in st.session_state:
+            agent_file = st.session_state.agent_file
+            workflow_file = st.session_state.workflow_file
+            ma_gent_workflow_workflow_ui = StreamlitWorkflowUI(agent_file, workflow_file, '', 'Maestro meta-agents generated workflow')
             ma_gent_workflow_workflow_ui.setup_ui()
 
 if __name__ == '__main__':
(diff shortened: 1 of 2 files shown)


akihikokuroda:
Mark as a draft waiting for #394

akihikokuroda:
@george-lhj I see.  Is it possible to make the agent to upload the files to a URL?  If so, we can create a tiny web server that receives the files and writes them in the specific location.

akihikokuroda:
The example command uploading file is 
```
curl -F "file=@agent.yaml"  http://localhost:5001/
```

akihikokuroda:
Here is an example of the server.
```
import os
from flask import Flask, flash, request, redirect, url_for
from werkzeug.utils import secure_filename

UPLOAD_FOLDER = './files'
ALLOWED_EXTENSIONS = {'yaml', 'txt'}

app = Flask(__name__)
app.config['UPLOAD_FOLDER'] = UPLOAD_FOLDER

def allowed_file(filename):
    return '.' in filename and \
           filename.rsplit('.', 1)[1].lower() in ALLOWED_EXTENSIONS

@app.route('/', methods=['POST'])
def upload_file():
    if request.method == 'POST':
        # check if the post request has the file part
        if 'file' not in request.files:
            flash('No file part')
            return redirect(request.url)
        file = request.files['file']
        # If the user does not select a file, the browser submits an
        # empty file without a filename.
        if file.filename == '':
            flash('No selected file')
            return redirect(request.url)
        if file and allowed_file(file.filename):
            filename = secure_filename(file.filename)
            file.save(os.path.join(app.config['UPLOAD_FOLDER'], filename))
            return '', 200
    return '', 400

if __name__ == '__main__':
    app.run(host='0.0.0.0', debug=True, port=5001)
```

maximilien (commented):
This is fine. I assume you are syncing with @george-lhj to add the generated files in the `sessions_state`?

akihikokuroda:
@george-lhj The other way is marking the file contents in the agent response like.
```
<file name=agent.yaml>
file contents.
</file>
```
So the file can be created in `StreamlitWorkflowUI`.

george-lhj:
Ok, I will see if the agent in bee is able to do this

george-lhj:
so after testing, curl commands are not supported, the LLMs we have access to cannot directly write anything nor have the ability to upload to a server (these are security issues that all the models have in place) so the alternative approach also doesn't work

george-lhj:
> @george-lhj The other way is marking the file contents in the agent response like.
> 
> ```
> <file name=agent.yaml>
> file contents.
> </file>
> ```
> 
> So the file can be created in `StreamlitWorkflowUI`.

so this approach might work, but not sure if we are thinking the same solutoin.

I can get the LLM to create a response with a tag. So say my input is the content of agent.yaml file, and then I pass this into the new agent. Then, I can get the agent to output something with specifc tag

so like what you said:

<file name=agent.yaml>
content
</file>

and then somewhere else we would have to add support to parse this?
Is this also what you were thinking? @akihikokuroda

george-lhj:
Screenshot 2025-04-08 at 2 09 37 PM
i can get the LLM to output something like this, but we cannot directly save,modify, upload anything related to files

akihikokuroda:
Good.  I can add the parsing code in the workflow UI to capture the contents in files.  Then making the Generated tab to load the files

george-lhj:
could you also make an option to upload the files you want directly? so lets say in case the ones that generated are wrong, we can edit it, save as some file and then reupload/retry

george-lhj:
you can base the output off #404, which was merged

maximilien:
LMK when this is ready for review since #404 is now merged.

akihikokuroda:
@george-lhj I couldn't make the meta agent to generate the yaml file contents in the response message.  I used some mock response to test this change.  Would you try?  The yaml contents are saved in `generated_agent` and `generated_workflow`.  The `load generated files` button in the `Generated` tab loads them.

george-lhj:
ok i will try

george-lhj (commented):
so I am able to successfully generate this output using the streamlit, as you can see here.
Screenshot 2025-04-09 at 10 42 16 AM

When I go to the generated tab and click the button, nothing seems to happen. Do I need to run this via CLI first?  And then load run the deploy/streamlit?

I sent a video to you on slack if you want to see it.

akihikokuroda:
@george-lhj Thanks!  Both agent and workflow yaml need to be generated to get them loaded.  Would you try?

george-lhj:
So I got them both generated but still seems like nothing came up. I did not reset/reload anything, just after the agent file generated I went to the next workflow section and generated it

akihikokuroda:
@george-lhj I hope I fixed the issue in the last commit.  Thanks!

george-lhj (approved):
im assuming the expected outcome is what i described in slack, and if so, we can merge

george-lhj:
this one is done, the load button loads the generated files (but alot of the time there is an issue with the generated yaml) so the user needs to do some small fixes

PR #404, opened 2025-04-08
Add parsers

### LINKED ISSUE #394: bug: meta-agents should generate agents/workflow file

george-lhj:
fixes #394 

add an agent to takes in the final yaml format of the output, and then adds tags

<file start>
content
</file>

so that we can parse the content and directly upload

here is what output looks like:

Screenshot 2025-04-08 at 2 50 47 PM


maximilien (approved):
I see so you are using tags to delineate the output. I still think a tool to write file should be possible. We can explore that option later

PR #408, opened 2025-04-10
Fix local beeai agent error

akihikokuroda:
The local beeai agent was broken.  This fix the issue.

### DIFF
--- a/maestro/src/agents/beeai_local_agent.py
+++ b/maestro/src/agents/beeai_local_agent.py
@@ -103,10 +103,12 @@ class BeeAILocalAgent(Agent):
         llm = OllamaChatModel(self.agent_model)
 
         templates: dict[str, Any] = {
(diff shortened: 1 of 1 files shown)


george-lhj (approved):

maximilien (commented):
Nice.

was thinking of a change to YAML as follows for specifying local or remote:

```yaml
apiVersion: maestro/v1alpha1
kind: Agent
metadata:
  name: beetest
  labels:
    app: testapp
spec:
  model: "llama3.1:latest"
  description: test
  instructions: input = 'Mock agent:'
  framework: bee
    mode: local
```

Of course, if mode is not specified then we can assume `local`. WDYT?

akihikokuroda:
I thought it is better if the agent can be switched between local and remote without changing the yaml file.  If it doesn't matter, I can update the agent schema and update code.

Issue #373, opened 2025-03-30
bug: fix what is causing the streamlit runtime warning


george-lhj:
`WARNING streamlit.runtime.scriptrunner_utils.script_run_context: Thread 'MainThread': missing ScriptRunContext! This warning can be ignored when running in bare mode.`

Appears in the terminal every time we run, even when we have closed streamlit/are not deploying

psschwei:
The issue seems to be that streamlit doesn't support multithreading, and you need to hack it in by following one of the options on this page: https://docs.streamlit.io/develop/concepts/design/multithreading

I don't think that's a trivial fix... it would probably require substantial rewrites to the code base.

The doc above also links to this page, which has some other ideas: https://sehmi-conscious.medium.com/got-that-asyncio-feeling-f1a7c37cab8b

There's a chance that [temporarily suppressing warnings](https://docs.python.org/3/library/warnings.html#temporarily-suppressing-warnings) might catch this, but that would block ALL warnings which we probably don't want. But might be acceptable for demos...

maximilien:
Closing this as we should create new Node.JS (et al.) default UI for workflows. 

The keys are:

1. It's native WASM 
2. JavaScript / TypeScript
3. Fast and supports the various Node libraries

maximilien:
Closing this.

PR #410, opened 2025-04-10
first attempt at solving issue #373 but does not work

maximilien:
This attempts to solve issue #373 but fails. doing PR just to share and so whoever spends time on this issue can reuse or at least don't spend time on this approach which seems should work.

### DIFF
--- a/maestro/cli/commands.py
+++ b/maestro/cli/commands.py
@@ -23,7 +23,7 @@ from jsonschema.exceptions import ValidationError, SchemaError
 
 from src.deploy import Deploy
 from src.workflow import Workflow, create_agents
-from cli.common import Console, parse_yaml, read_file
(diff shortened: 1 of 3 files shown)


maximilien:
DON'T MERGE @akihikokuroda since does not work.

maximilien:
This is the code for ease of finding it:

```python
# workaround to remove streamlit WARNING:
# WARNING streamlit.runtime.scriptrunner_utils.script_run_context: Thread 'MainThread': missing ScriptRunContext!
def remove_streamlit_logging_warnings():
    import logging
    for name, l in logging.root.manager.loggerDict.items():
        if "streamlit" in name:
            l.disabled = True
```

and just execute `remove_streamlit_logging_warnings()` before launching `streamlit`.

However as I said this does not work.

Issue #407, opened 2025-04-10
bug: streamlit_deploy not found (external usage)


planetf1:
When working on an independent project with maestro as a dependency (`requirements.txt`):
```
maestro @ git+https://github.com/i-am-bee/beeai-labs#subdirectory=maestro
```
```
((.venv) ) ➜  remediation-new git:(dev/jonesn/remediation-new) ✗ maestro deploy agents.yaml workflow.yaml --streamlit
2025-04-09 18:49:52.948 WARNING streamlit.runtime.scriptrunner_utils.script_run_context: Thread 'MainThread': missing ScriptRunContext! This warning can be ignored when running in bare mode.
Workflow deployed: http://localhost:8501/?embed=true
Usage: streamlit run [OPTIONS] TARGET [ARGS]...
Try 'streamlit run --help' for help.

Error: Invalid value: File does not exist: ./cli/streamlit_deploy.py
```

The *warning* is covered by issue #373

However there's an additional, different issue here with accessing streamlit_deploy. This issue targets that problem.

maximilien:
This is an easy fix @planetf1 by essentially passing the directory where the maestro stript is. I'll fix it.

PR #411, opened 2025-04-10
fixes issue #407

maximilien:


maximilien:
But it's OK to have it run this way for now.

Issue #384, opened 2025-04-01
feature: add other agents


maximilien:
Support for other agents:

- [x] streamline BeeAI (local and remote) -- should be one YAML with framework and mode (local or remote) and factory can use that to create the right instance
- [x] tests for BeeAI (local and remote)
- [x] test CrewAI agents
- [ ] ADK agents
- [ ] smol agents
- [ ] test ADK agents
- [ ] test smol agents
- [ ] test workflow with all the supported agents (issue #387)

@akihikokuroda feel free to help here. Just PR against this issue. I am working on the OpenAI agents support. So all others are free to take now.

akihikokuroda:
I'm working on
> streamline BeeAI (local and remote) -- should be one YAML with framework and mode (local or remote) and factory can use that to create the right instance

Is it ok to switch between local and remote agents using the environment variable `LOCAL_AGENT`? @maximilien

maximilien:
Sounds good on end variable. I marked the first sub task as done @akihikokuroda. Feel free to start checking tasks off as you submit PRs.

PR #413, opened 2025-04-10
Streamline remote/local agent

akihikokuroda:
fix 
>streamline BeeAI (local and remote) -- should be one YAML with framework and mode (local or remote) and factory can use that to create the right instance
in #384 

Setting "remote" in `spec.mode` uses the "remote" agent.  The default (`spec.mode` is not set) is the local agent.


maximilien (approved):
Thank you @akihikokuroda this looks good. I will have to do a big merge with my draft code but this was important to do.

LGTM

PR #416, opened 2025-04-11
updating summary agents

george-lhj:
since we are demoing the summary agents, i further updated the entire demo to be a bit more consistent. 
updated readme also

created 2 new agents that replacted the old ones:

one that retrieves abstract from the arxiv based on the fetched titles
one that generates summary based on the paper and abstract


george-lhj (commented):
note: need to change all the past demo to have remote in it, it is outdated

maximilien (approved):
Can we have a video of this. Also, I want to see if we can add some multimodality into our demos. So for this one I wonder if we could include a word cloud as a visual for the final summary. In streamlit that can show nicely since the output is markdown.

LGTM

maximilien (commented):
LGTM

PR #419, opened 2025-04-14
bug: update agent defs

### LINKED ISSUE #417: bug: update all agents demos to match latest PR
After the latest PR, the schema requires agents to have mode and framework under spec in the agents.yaml defintion. this needs to be added, otherwise the creation `maestro create agents.yaml` will fail.

george-lhj:
agent creation got updated, so need to update the meta agent prompts to include these required tags.

Also updated agents.yaml files for each demo to reflect this change as well

fixes #417


george-lhj:
@akihikokuroda  any idea why the tests are all failing? seems like your previous PR when i approved also passed the tests, but after merge it is failing.

Is it a gh issue?

akihikokuroda:
It's caused by the beeai_framework change.  I'll fix it.

akihikokuroda (approved):
LGTM!!

PR #421, opened 2025-04-14
demo/enhancement: Update summary arxiv agent

george-lhj:
Previously, using a custom tool to fetch from arxiv. Now simplifying the demo through the usage of the arxiv tool directly to pull n papers related to some topic. 

+ a few prompt changes

will be using this version to record demo


maximilien (approved):
LGTM

PR #423, opened 2025-04-14
demo: update agents defs

george-lhj:
further editing to output in a markdown format for better readability


maximilien (commented):
Nice. However, where in the instructions is it told to "read" the paper? It seems like it's only the abstract? Am I wrong?

PR #426, opened 2025-04-15
Add module, class and factory in meta.labels

akihikokuroda:
Move "module", "class" and "factory" in `meta.name` to `meta.labels` so that the name can have any string


george-lhj (approved):

maximilien (approved):
LGTM

Issue #350, opened 2025-03-25
enhancement: remove need to pass schema file when validating `agents.yaml` or `workflow.yaml`


maximilien:
Since the `agents.yaml` and `workflow.yaml` can be inferred to be agents or workflow file with type and structure the CLI should be simplified and download the schema files from source:

That way user can validate with:`maestro validate agents.yaml` and `maestro validate workflow.yaml`

This is nice to have and not urgent.

PR #429, opened 2025-04-15
feature: fixes issue #350

maximilien:
fixes issue #350 Now the maestro CLI can be called to validate w/o passing the schema file since this can be discovered from the YAML file. 

This PR works for both agents and workflows and even tools (which is not yet supported). You can test this with: `maestro validate tests/yamls/agents/simple_agent.yaml`

This also changes the `tools/check-schemas.sh` to use the new fashion. However, there is a CrewAI agent config that fails, but somehow did not before... so @akihikokuroda I did that change as well

When done this will complete issue #350


akihikokuroda (approved):
LGTM

PR #433, opened 2025-04-17
enhancement: prompt adjustment

george-lhj:
Weather checker prompt doesn't perform super consistently, so giving an example process allows it to act faster and more accurate.


maximilien (approved):
LGTM

PR #434, opened 2025-04-17
litellm dependency conflict issue workaround

akihikokuroda:
This is workaround for the litellm dependency issue between beeai_framework and crewai

Crewai is required older version of the litellm.  It be updated to resolve this issue.

### DIFF
--- a/.github/workflows/maestro_check-mermaid.yaml
+++ b/.github/workflows/maestro_check-mermaid.yaml
@@ -27,8 +27,7 @@ jobs:
         run: |
           cd maestro
           poetry install
-          poetry run pip install crewai
-          poetry run pip list
+          poetry run pip install beeai-framework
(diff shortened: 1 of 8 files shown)


george-lhj (approved):

george-lhj:
lgtm

maximilien (approved):
Interesting so is the workaround to install beeai-framework only in the GH actions. Can you explain the details so it's captured. Thanks.

akihikokuroda:
The issue that this PR workaround is 

> ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
crewai 0.114.0 requires litellm==1.60.2, but you have litellm 1.66.3 which is incompatible.

The `litellm 1.66.3` is required by beeai-framework.

PR #448, opened 2025-04-23
Add MCP tool support

planetf1:
Adds support for MCP tools (first part)
- Currently this is implemented as a new module which could be shareable between agent backends.
- For now it does also contain some OpenAI type tool support (but other llm backends may need this)
- A test is included, but will not be run automatically as depends on having an MCP SSE server active
- The OpenAI agent run method will initiatialise the tool list

In progress/todo
 - Ensuring OpenAI agent can call a tool 
 - clean up output/debug 

Future PR:
 - Update/add schemas / deployment info / environment variables to allow the specification of MCP endpoints & any prompt changes etc

### DIFF
--- a/maestro/pyproject.toml
+++ b/maestro/pyproject.toml
@@ -26,6 +26,8 @@ openapi = "^2.0.0"
 openai-agents = "^0.0.11"
 pycron = "^3.1.2"
 beeai-framework = "^0.1.17"
+pytest-asyncio = "^0.26.0"
+mcp = "^1.6.0"
 
 [tool.poetry.group.dev.dependencies]
 black = "^24.10.0"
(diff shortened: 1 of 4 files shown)


planetf1:
Will squash/rebase once prior PRs merged (tomorrow)

planetf1:
Here's an example of the test which shows a 'remote' (actually localhost) MCP server returning a list of tools, and then the test case calling the add function.

[log.txt](https://github.com/user-attachments/files/19872931/log.txt)

planetf1:
Closing for now.

The standalone MCP code works, but it's not aligned with the Agents SDK - so the implementation needs a re-think. See issue for further information

PR #457, opened 2025-04-27
demo: new summary demo to align w/clients needs

george-lhj:
fully redesigned summary ai using agents in order to support the following client needs:

1) Know when new IBM papers hit the ArXiv
2) Search ArXiv for papers on a certain topic during a certain time period (only for IBM documents)
3) Retreive metadata, create summary for certain paper
4) Send the summary to your slack

future updates:
1A) Look into creating own tool in python that can create better summary from PDF
1B) Currently only output of first agent is something like this ['title1','title2', etc]. Currently the rest of the agents only can handle one title at a time, but I need support for a loop or parallel with each title in output @akihikokuroda 
...
2) Support for non ibm docs? (could be a lot slower because the freuqency of released docs are much higher)

### DIFF
--- a/maestro/demos/README.md
+++ b/maestro/demos/README.md
@@ -87,7 +87,7 @@ curl -X POST http://localhost:11434/api/generate -d '{
 
 #### Allowing maestro to be run from anywhere
 
-Modify wrapper script: `nano ~/.local/bin/maestro`
+Create wrapper script: `nano ~/.local/bin/maestro`
 Set the script path to run relative to your location, whereever in the terminal:
 
 ```bash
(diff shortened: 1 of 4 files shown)


george-lhj:
@psschwei could you take a look at this when you have the chance since both max and aki are out this week?

psschwei (commented):
I wasn't able to actually deploy the agents for this demo, so that's as far as I was able to get. Though I'll also note without comment that it doesn't seem that y'all are requiring approvals before PRs can merge ;)

george-lhj (commented):
thanks

george-lhj:
> Should this just be another demo (different than initial summary) so that way we have two

Sure, I can make it/refactor

PR #462, opened 2025-05-05
enhancement: generalize prompt

george-lhj:
New updates w more generalized instructions to allow agents to pull from differerent categories in arXiv.

Format updates to output to make it more readable in the summary.


george-lhj:
note: the summary can be even further enhanced usnig the openai agents in the future

PR #464, opened 2025-05-06
demo: summary.ai/ibm-summary.ai 

george-lhj:
Improved old version of summary ai, and also separated the specific IBM paper related demo that was requested.

Summary.ai now uses its own arxiv tool, which retreives papers on a specific subject (the n most recent papers), and then pulls its metadata, and creates summary which gets sent to slack. 

IBM version one has a differing agent that naively checks for possible IBM papers that were released to arxiv without accessing the employee database, and then does a similar process.


maestro/demos/workflows/ibm-summary.ai/agents.yaml
```diff
@@ -0,0 +1,174 @@
...
+  description: "Retrieve IBM related papers from a certain topic in a certain time frame"
+  instructions: |
+    Input:
+          • category (string) # e.g. "quantum-ph"
+          • since_days (int)  # e.g. "1" for the last 1 days since today's date.
+
+        Task:
+          1. Call the ibm-arXiv tool to search for all papers on `category` published in the last `since_days`. All the functionality is already implemented in the funciton, simply execute the function with the given parameters as the input, and then return the output (which should be titles of research papers) into a list.
+          Directly print out the list in the output.
+  code: |
```
maximilien, line 35: One of the things we need to work on is simplifying how these custom tools are added ...

maximilien (approved):
Love this @george-lhj but the proof will be in users :)

Also, we need to add an issue on simplifying how custom tools are added to agents. Let me look and create new issue (if one does not exist) and we can discuss details there.

Thanks for this.

PR #470, opened 2025-05-07
enhancement: added nlp agent for summary 

### LINKED ISSUE #465: demo: add agent to handle inital prompt
currently the initial prompt is like a funciton call: function(x, x).

need to add an initial agent that can take in either this function call, or use natural language description that would result in the function call also.

- from feedback in meeting

george-lhj:
fixes #465
previously input prompt was something like this: function_name(param). now, added support via intial agent that can take in the english description to generate and output the correct function call

### DIFF
--- a/demos/workflows/summary.ai/README.md
+++ b/demos/workflows/summary.ai/README.md
@@ -7,14 +7,16 @@ A multi-agent workflow using Maestro: Allows user to retrieve the latests ArXiV 
 <!-- MERMAID_START -->
 ```mermaid
 sequenceDiagram
+participant Query Interpreter
 participant All Papers
 participant get metadata
 participant generate summary
 participant slack
-All Papers->>get metadata: Step1
-get metadata->>generate summary: Step2
-generate summary->>slack: Step3
-slack->>slack: Step4
+Query Interpreter->>All Papers: Step1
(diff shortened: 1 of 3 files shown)


demos/workflows/summary.ai/README.md
```diff
@@ -7,14 +7,16 @@ A multi-agent workflow using Maestro: Allows user to retrieve the latests ArXiV
 <!-- MERMAID_START -->
 ```mermaid
 sequenceDiagram
+participant Query Interpreter
 participant All Papers
 participant get metadata
 participant generate summary
 participant slack
-All Papers->>get metadata: Step1
```
maximilien, line 14: I like naming the steps as `Step<n>` but only when it's a sequential step by step workflow... which I assume is the case here? Or asked differently, are there better names for `Step<n>`?

maximilien (approved):
Love this. Left a small comment. If needed address (with rename) or leave as is.

demos/workflows/summary.ai/README.md
george-lhj: This is a sequential so in the mermaid I think it makes sense for the reader. In the future I want to be able to run steps 2-4 in parallel based on the output of step 1, so if it makes sense then I will rename

PR #473, opened 2025-05-08
add crewai to requirements.txt

maximilien:
OK see if this simple fix works for you @george-lhj and @akihikokuroda


george-lhj (commented):
this one did not work directly for me. When I installed the requirements file I found our pyproject.toml has conflicting versions.
for example:
```     
The user requested psutil==7.0.0
The user requested psutil==5.9.8
```

I can look at this after lunch

george-lhj:
@akihikokuroda do you get the same as me?

george-lhj:
IMO I think the requirements file currently is outdated (and unnecessary since we have added poetry), if we are making it change to add crewai support I think it should be in the `pyproject.toml` file

george-lhj:
I tested adding it to the poetry file, but `crewai@1.67.0.post1` doesn’t exist on the default index, Poetry won’t resolve it which is the reason Aki and NIgel added the manual step at the end. We can add this to the requirements.txt file, but need to remove all of the clashing dependencies

PR #475, opened 2025-05-08
bug: Fix dependency installing warning

george-lhj:
Fixes #473 

We don't use requirements.txt to install the dependencies so removed the file. Created a script/command that user can use directly in order to install necessary dependcies from scratch and activate directly in the venv so that user can use maestro.


george-lhj:
@maximilien this should fix the warning we are having

maximilien (commented):
Love it. Thanks @george-lhj

Issue #476, opened 2025-05-08
chore: various clean ups


maximilien:
Let's use this issue to clean up things in maestro. Especially removing old beeai-labs files and dependencies -- unless needed for BeeAI framework support.

I also have a couple of new maestro images to add.

maximilien:
PR https://github.com/AI4quantum/maestro/pull/539

PR #477, opened 2025-05-08
remove old .env.example and add new images

maximilien:
@george-lhj, @planetf1, and @akihikokuroda let's use issue #476 for clean ups and then we can close once we are good.

Feel free to merge this this for me since it's straighforward


maximilien:
And if you like the Gemini logos @george-lhj we can change the streamlit UI to use one of them. Cheers.

images/maestro-gemini2.png
george-lhj, line 1: how come there is 2 agents.yaml for this image?

PR #479, opened 2025-05-09
enhancement: update actions file w setup

george-lhj:
Ensured setup for each run is the same as local virtual env by following the `setup_maestro` script

### DIFF
--- a/.github/workflows/maestro_demo-tests.yaml
+++ b/.github/workflows/maestro_demo-tests.yaml
@@ -3,40 +3,38 @@ on:
   push:
     branches: [ 'main' ]
     paths:
-    - '**'
+      - '**'
   pull_request:
     branches: [ 'main' ]
     paths:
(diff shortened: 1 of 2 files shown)


.github/workflows/maestro_demo-tests.yaml
```diff
@@ -3,40 +3,39 @@ on:
...
         run: |
-          poetry install
-          poetry run pip install crewai
-      - name: Verify Maestro Installation
-        run: |
-          poetry run maestro --help
+          chmod +x ./tools/setup_maestro.sh
+          source ./tools/setup_maestro.sh
+          which maestro
+          maestro --help
```
akihikokuroda, line 33: Are these for debugging?  Would you take them out?

akihikokuroda (approved):
LGTM!  I have a nit comment.

PR #480, opened 2025-05-12
feature: adding metric evaluation agent using opik

george-lhj:
For LLM evaluation, user want to rank/determine the best performing responses. OPIK is open source and does it, but makes calls from openai by default. As a workaround i have added support in the src file for this agent to point it at the ollama backend that we are running.


akihikokuroda (commented):
The metrics agent looks good.

maximilien:
Opik supports other LLMs and frameworks

george-lhj:
@akihikokuroda I have finished implementing the metrics. had to add some workarounds inside the `metrics_agent.py` because litellm and opik by default both use OpenAI. I have to set it to be in the ollama endpoint

maximilien (commented):
Love it. Need to test this. Do we have any demos using yet?

maximilien (approved):
👍🏽

george-lhj:
not yet, planning to add this into our demos once i update

PR #488, opened 2025-05-22
chore: various cleanups mosty using PyLint suggestions

maximilien:
Clean ups of the `maestro/cli` module using PyLint


george-lhj (approved):
is there a specific linter/linting format you want us to use or follow?

PR #497, opened 2025-06-02
CBOM.ai : port to openai backend

planetf1:
This PR 'ports' the cbom.ai demo to run under the 'openai' backend.

This runs end-end with no cut/pasting of code or manual enabling of tools (needed for deprecated remote backend)

* Models remain the same
* prompts - minor changes to specify the tool name as the function name
* copied the two python functions into standalone files (original inline code in yaml left intact, but has no impact)
* custom agent which echoed a single line (part of maestro stack) replaced with a regular agent

To run, this requires the two custom functions to be available via MCP.

An *example* is to configure the openai support in maestro 
1) Get an example mcp server https://github.com/planetf1/mcp-server
2) Set the maestro environment to create a mcp server instance based upon the two custom functions:
```
export MAESTRO_MCP_ENDPOINTS="python3 /Users/jonesn/AI/mcp-server/mcp_server.py --log /tmp/mcp.log demos/workflows/cbom.ai/java_fetcher.py demos/workflows/cbom.ai/fetch_code.py"
```
Note: A log of all tool calls is appended to /tmp/mcp.log

FYI @george-lhj


planetf1:
I created #503 with new filenames (as it also includes the (not working) attempt to include the additional remediation steps).

Please take that one if you prefer (just one)

I'll update the paths here too so you can chose (but without the additional tools/readme etc)

PR #498, opened 2025-06-02
bug: scoring agent default model

george-lhj:
Currently the litellm default model is set to openai, which doesn't work unless we have an openai key. This is causing issues when opik calls evaluation as it requires LLM calling.

Switching the backend to be ollama which works.


george-lhj:
@akihikokuroda i cannot request a review, but this is a bug i found when reviewing the newest PR

PR #501, opened 2025-06-03
add CodeAgent

### LINKED ISSUE #499: feature: add a CodeAgent for simple code execution
This agent would not use an LLM and simply execute code specified.

This is useful when LLM or memory is not required and the work to be done can be specified procedurally and never changes, e.g., Git workflow.

This came up during cbom.ai demo implementation where git manipulation is needed and needs to be 100% the same every time but with different inputs.

Cc: @george-lhj @planetf1

akihikokuroda:
fix #499

### DIFF
--- a/src/agents/code_agent.py
+++ b/src/agents/code_agent.py
@@ -1,0 +1,53 @@
+#! /usr/bin/env python3
+# SPDX-License-Identifier: Apache-2.0
+
+import asyncio
+import os
+from dotenv import load_dotenv
+from slack_sdk import WebClient
+from slack_sdk.errors import SlackApiError
+
+from src.agents.agent import Agent
+
+load_dotenv()
+
+class CodeAgent(Agent):
+    """
+    CodeAgent extends the Agent class that executes an arbitrary python code specifed in the code section of the agent difinition.
+    """
+
+    def __init__(self, agent: dict) -> None:
+        """
+        Initializes the agent with agent definitions.
+        """
+        super().__init__(agent)
+        
+
+    async def run(self, prompt: str) -> str:
+        """
+        Runs the BeeAI agent with the given prompt.
+        Args:
+            prompt (str): The prompt to run the agent with.
+        """
+
+        self.print(f"Running {self.agent_name} with {prompt}...\n")
+        local = {"input": prompt}
+        exec(self.agent_code, local)
+        answer = local["input"]
+        self.print(f"Response from {self.agent_name}: {answer}\n")        
+        return local["input"]
+
+    async def run_streaming(self, prompt: str) -> str:
+        """
+        Runs the BeeAI agent with the given prompt.
+        Args:
+            prompt (str): The prompt to run the agent with.
+        """
+
+        self.print(f"Running {self.agent_name} with {prompt}...\n")
+        local = {"input": prompt}
+        exec(self.agent_code, local)
+        answer = local["input"]
+        self.print(f"Response from {self.agent_name}: {answer}\n")        
+        return local["input"]
+
--- a/tests/yamls/agents/code_agent.yaml
+++ b/tests/yamls/agents/code_agent.yaml
@@ -1,0 +1,16 @@
+apiVersion: maestro/v1alpha1
+kind: Agent
+metadata:
+  name: code
+  labels:
+    app: slack-example
+spec:
+  model: dummy
+  framework: code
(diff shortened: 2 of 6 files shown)


src/agents/code_agent.py
```diff
@@ -0,0 +1,53 @@
...
+    """
+
+    def __init__(self, agent: dict) -> None:
+        """
+        Initializes the agent with agent definitions.
+        """
+        super().__init__(agent)
+        
+
+    async def run(self, prompt: str) -> str:
```
george-lhj, line 26: I remember that I changed the input for run to be a dictionary so that it can accept context as well, I think perhaps you need to edit this one so that we can support it as well.

george-lhj: see the tag I tagged you in, also can find it here:
https://github.com/AI4quantum/maestro/pull/493/files under step.py, the run method

src/agents/code_agent.py
```diff
@@ -0,0 +1,53 @@
...
+from slack_sdk import WebClient
+from slack_sdk.errors import SlackApiError
+
+from src.agents.agent import Agent
+
+load_dotenv()
+
+class CodeAgent(Agent):
+    """
+    CodeAgent extends the Agent class that executes an arbitrary python code specifed in the code section of the agent difinition.
```
maximilien, line 16: s/difinition/definition/

src/agents/code_agent.py
```diff
@@ -0,0 +1,53 @@
...
+    def __init__(self, agent: dict) -> None:
+        """
+        Initializes the agent with agent definitions.
+        """
+        super().__init__(agent)
+        
+
+    async def run(self, *args, context=None) -> str:
+        """
+        Runs the BeeAI agent with the given prompt.
```
maximilien, line 28: remove BeeAI agent reference

tests/yamls/agents/code_agent.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: maestro/v1alpha1
+kind: Agent
+metadata:
+  name: code
+  labels:
+    app: slack-example
+spec:
+  model: dummy
```
maximilien, line 8: shouldn't we just not specify model?

tests/yamls/agents/code_agent.yaml
```diff
@@ -0,0 +1,16 @@
...
+  framework: code
```
maximilien, line 9: love this `code` as framework. Maybe we should have `language: python` since we could in theory have javascript? but OK to leave that as comment for future.

tests/yamls/agents/code_agent.yaml
```diff
@@ -0,0 +1,16 @@
...
+  mode: local
+  description: code agnet
```
maximilien, line 11: typo

src/agents/code_agent.py
```diff
@@ -0,0 +1,53 @@
...
+        Args:
+            prompt (str): The prompt to run the agent with.
+        """
+
+        self.print(f"Running {self.agent_name} with {args}...\n")
+        local = {"input": args}
+        exec(self.agent_code, local)
```
maximilien, line 35: what happens if code fails. We should try except and print log and check for exception step

maximilien (changes requested):
Good start. Left various feedback.

george-lhj:
From discussion w/ @maximilien , it sounds like the CodeAgent is intended to execute arbitrary Python code, potentially allowing things like defining and running functions (e.g., creating a GitHub PR). Correct me if I am wrong.

Reading the current implementation, it looks like the executed code must modify the input variable directly to return anything meaningful. This seems pretty limited in practice and might not support more realistic workflows where we define and call functions, use imports, and return results via a separate output variable.

Would it make sense to support a more general execution pattern; e.g., executing a full block of code and returning a variable like output? That might better align with the goal of running arbitrary Python logic inside an agent.

akihikokuroda:
@george-lhj It can execute a full block of the code.  An empty dict for the output can be added.  The code can put the output in the dict.  It probably needs to be convert to a string.   WDYT?

george-lhj (commented):
going to review this within cbom.ai demo. probably going to add just a small test to see the code functionality because the writing the entire automation for gh pr might take too long

george-lhj:
thanks Aki, lgtm!

PR #503, opened 2025-06-04
CBOM.ai : port to openai backend (alternative to #497)

planetf1:
Signed-off-by: Nigel Jones <jonesn@uk.ibm.com>

@george-lhj As discussed here's an updated version which
 - creates ported workflow with alternate file names
 - port of old agents added - not working, commented out
 
 Is this helpful for you to continue working? If so feel free to merge (and close #497)


george-lhj (commented):
I'm able to reproduce the output you are seeing.

However, I still think its important at least for now to separate the remediation files w/the original workflow, since they have separate processes in terms of tool calling (with the additional py files), setup, etc. Could you create a separate remediation folder inside cbom.ai that moves this process into one folder? And then keep the original bee-ai framework files?

PR #504, opened 2025-06-04
enhance step/inputs

### LINKED ISSUE #500: nice to have: Scoring Agent Instruction Reference in Workflow Input
Currently, Maestro supports the `inputs.from` field in a workflow step as one of:
- `"prompt"` — to use the global template prompt
- A step name — to use the output of a previous agent
- A string literal — manually written by the user

However, **referencing the instructions of an agent used in a previous step is not yet supported**. This makes it difficult to dynamically retrieve the intent of a step during evaluation or scoring workflows. Instead we bypass with string literals to move fast under time pressure.

---

### Use Case
In evaluation or scoring workflows, the evaluator agent should understand:
1. What the previous agent *was supposed to do* (its instructions)
2. What it actually *did* (its output)

Right now, users must manually copy the instructions from `agents.yaml` into a string literal in the workflow. This is repetitive and error-prone.

---

### Desired Syntax

```yaml
inputs:
  - from: instructions:step3
  - from: step3
```

This should:

1. Resolve step3 to its agent name (e.g., raw gh reader)
2. Look up that agent in the loaded agents.yaml
3. Retrieve the value under spec.instructions
4. Pass it as the input to the current step

#### Proposed Implementation
In the Workflow class:
Build a dictionary agent_def_map from agent name → full definition in __init__()
During input resolution in _condition() and _condition_subflow():
Detect from: instructions:stepName
Resolve and return the corresponding agent’s instructions string

##### Status
String literal support is already implemented and sufficient for current use. This enhancement is deferred for future implementation.

akihikokuroda:
fix #500 

This PR enables a new input type: pulling an agent’s instructions field from a previous step (e.g., for reflection, reuse, meta-programming, etc.).

### DIFF
--- a/src/workflow.py
+++ b/src/workflow.py
@@ -137,6 +137,8 @@ class Workflow:
                     src = inp["from"]
                     if src == "prompt":
                         args.append(initial_prompt)
+                    elif "instructions:" in src:
+                        args.append(step_defs[src.split(":")[-1]]["agent"].agent_instr)
                     elif src in step_results:
                         args.append(step_results[src])
                     else:
--- a/tests/workflow/test_inputs.py
+++ b/tests/workflow/test_inputs.py
@@ -1,0 +1,47 @@
+#!/usr/bin/env python3
+
+# Copyright © 2025 IBM
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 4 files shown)


george-lhj (approved):
thanks lgtm

PR #512, opened 2025-06-10
demo: change fetch tool from llm to codeagent

### LINKED ISSUE #511: task: change git-fetcher to a code agent
Currently to get the java files inside of cbom workflow, we are using LLM which is taking ~5 minutes. To reduce the waiting time, change implementation to a code agent.

george-lhj:
fixes #511, reduces time from 5 min to <1 sec for the initial agent.

@akihikokuroda am i correct in assuming since this agent is now a code agent, it should not be included in the `workflowrun.yaml` file? Since the other code agents are also not included?

### DIFF
--- a/demos/workflows/cbom.ai/agents.yaml
+++ b/demos/workflows/cbom.ai/agents.yaml
@@ -1,58 +1,31 @@
 apiVersion: maestro/v1alpha1
 kind: Agent
 metadata:
-  name: git fetcher
+  name: gh fetch
   labels:
     app: cbom.ai
 spec:
-  model: qwen3:latest
-  description: |
-    output the java files given a repository
-  instructions: |
-    You are the **Java Fetcher** agent.  
-    **IMPORTANT**: After you invoke the `java-fetcher` tool, you MUST **directly return** its output — exactly character-for-character—and nothing else.  
-    Do **not**:
-      - Reformat or prettify the list  
-      - Edit, Sort, filter, dedupe, or correct paths  
-      - Append any commentary  
-    **Steps:**  
-      1. Invoke the tool:  
-        ```Python
-        java-fetcher(owner="given owner input", repo="given repo input")
-        ```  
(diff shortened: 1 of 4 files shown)


george-lhj (commented):
for some reason, i cannot request you to review @maximilien but this is ready

demos/workflows/cbom.ai/agents.yaml
```diff
@@ -1,58 +1,31 @@
...
-      1. Invoke the tool:  
-        ```Python
-        java-fetcher(owner="given owner input", repo="given repo input")
-        ```  
-      3. Receive the tool’s response, an array of URLs: ['url1', 'url2', 'url3', etc ...]
-      4. **Immediately** print/output that array exactly as-is. *NOTHING ELSE AND NO EDITING*
-  mode: remote
-  tools:
-    - java-fetcher
+  framework: code
```
maximilien, line 8: love the simplicity now

demos/workflows/cbom.ai/workflow.yaml
```diff
@@ -41,7 +41,7 @@ spec:
         - from: "Given a raw github link, output its raw file directly."
         - from: step3
       context:
-        - "A correct output should just be the raw version of a java file."
+        - "The current output should be the following: https://raw.githubusercontent.com/Mastercard/client-encryption-java/main/src/main/java/com/mastercard/developer/encryption/EncryptionConfig.java"
```
maximilien, line 44: why do you need this? And if we do then perhaps a Java file from a more stable Java repository would make sense? WDYT?

maximilien (approved):
Love this. Thanks for small PR too :)

demos/workflows/cbom.ai/workflow.yaml
george-lhj: This is purely for the demo, I just want to make sure that we are showcasing the scoring because I know the agent is returning that URL.

When we expand on this in the future, def agree w you

PR #518, opened 2025-06-11
feature: add brew support for maestro

maximilien:
First brew formulae for maestro


maximilien:
This is first step. Need to work on creating checksum from release and figuring out which tap. Also, need to add a test. These will all come from separate PRs.

CC: @psschwei @ajbozarth

maximilien:
Closing this since pip is a solution that can work for more. If pip solution is not good enough then let’s re-explore this one.

PR #522, opened 2025-06-12
bug: values of metadata.labels must be sanitized

akihikokuroda:
values of metadata.labels must be sanitized to be in kubernates CR.

Add create-cr command error messages.


akihikokuroda:
One yaml file used in the "Operator Test" action is updated to fail if it is not sanitized.

maximilien (approved):
lgtm

PR #525, opened 2025-06-12
enhancement : Add release workflow

ajbozarth:


maximilien:
Tests failing @ajbozarth

ajbozarth:
> Tests failing

looked into this and it's because I added crewai  to the dependencies to remove the need for the setup_maestro script so when the workflow runs the setup_maestro script it breaks, given the work in #528 I believe it's better to wait for that work and clean up this PR afterwards. For now I'll be testing that maestro still works as intended when pip installing the distribution in a clean env

maximilien:
@ajbozarth why not close this and open a new one that includes the us changes. WDYT?

maximilien:
Why don't we close this Alex since new PR will be very different... note also you need to rebase since @george-lhj had a PR today that fix an important issue with the uv change that prevented streamlit deploys. Thx

ajbozarth:
Closing, I had hoped to just rebase but that didn't work. I'll submit a fresh PR tomorrow

PR #528, opened 2025-06-13
chore: massive PR to move from `poetry` to `uv`

maximilien:
While BIG (lots of changes and moves) both pytest and ./tools/run-all.sh are green locally. Assuming this works for you all then I we should be good to go :)

I have updated all documentations but it's entirely possible that more needs to be added especially in the demos/*/README.md files. So if you find any then suggest and I can change or better let's merge and do this in a separate PR -- especially if you are the owner of the demo and want to improve README.md.

# Guide to check this PR

## pip

First install `pip` if you don't have and also upgrade:

`pip install --upgrade pip`

## uv

Second, install `uv` if you haven't already:

`pip install uv`

## venv

Create a new virtual environment and install dependencies:

```bash
uv venv
source .venv/bin/activate  # On Linix/macOS
uv pip install -r requirements.txt
```

## Tests 

`uv run pytest`

and the tools/check*.sh

`uv run ./tools/run-all.sh`

## Development

For development dependencies, you can install them separately:

`uv pip install [python-package]`

## Notes

The main differences you'll notice when using uv instead of Poetry:

uv is significantly faster for dependency resolution and installation

Instead of `poetry add`, you'll use `uv pip install`

Instead of `poetry run`, you'll use the Python interpreter directly from your virtual environment using `uv run maestro ...`

CC: @AngeloDanducci @ajbozarth @psschwei @planetf1


.github/workflows/maestro_operator-test.yaml
```diff
@@ -74,14 +82,35 @@ jobs:
...
-    - name: Send request
+    - name: Wait for pod to be ready
       run: |
-        curl -L "http://127.0.0.1:30051/?Prompt=Los+angels"
+        kubectl wait --for=condition=ready pod -l app.kubernetes.io/instance=weather-checker-ai --timeout=60s
 
+    - name: Port forward and send request
+      run: |
+        # Start port-forward in background
+        kubectl port-forward svc/weather-checker-ai 30051:80 &
```
psschwei, line 92: In past experience, I've never had any luck trying to port-forward in a github action. Let me see what we used in serverless instead...

psschwei: we hacked the svc to make it a load balancer and then directly hit that

```
kubectl patch svc gateway -p '{"spec": {"type": "LoadBalancer"}}'
export GATEWAY_HOST="http://$(kubectl get svc gateway -o jsonpath="{.status.loadBalancer.ingress[0].ip}"):8000"
```

we then used `$GATEWAY_HOST` when we had to send API requests

maximilien: interesting. Let me try one more thing since Aki had it working before.

maximilien: OK trying your suggestion.

maximilien:
> As you're iterating on this, based on my work on #525 I believe we can just add `crewai = "^0.130.0"` to the dependencies in the `pyproject.toml` and get rid of the setup_maestro script, the issue that was causing that need has since been addressed upstream

Nope. Lots of changes but all green now. Except Operator Tests. Will need @akihikokuroda to fix that. Perhaps @psschwei can as well. But I am giving up on that one :)

george-lhj (commented):
Currently, after following the README installation steps, maestro isn’t immediately available as a CLI command. We have to manually run `uv pip install -e .` in order to register the entry point and make `maestro --help` work.

Would it make sense to include this directly as part of the requirements.txt or automate it in the current setup flow? That way, users would be able to run maestro immediately after following the installation instructions, without needing the extra manual step.

Note on missing dependencies:

Also, the CrewAI integration failure isn’t handled gracefully. Right now, users have to manually run: `uv pip install crewai litellm==1.67.0.post1`
to resolve the missing import. It might be helpful to either (1) include this in the installation instructions, or (2) handle the missing dependency more cleanly in code to avoid confusion.

akihikokuroda:
I'll work the container build in a follow up PR.

maximilien:
If @akihikokuroda is OK. Let's merge and fix different smaller issues separately... but don't let me influence your reviews ;)

akihikokuroda (approved):
LGTM.  Any issues can be in follow up PRs.

PR #530, opened 2025-06-14
Dependency clash

### LINKED ISSUE #529: bug: scoring agent string error after uv merge

george-lhj:
fixes #529 

During testing, found that ScoringAgent attempt to parse the litellm version using a strict SemVer parser.

When installing litellm==1.67.0.post1 (which is valid per PEP 440 and installed via uv), the version string 1.67.0.post1 is not valid SemVer and causes runtime parsing errors like:

'1.67.0.post1' is not valid SemVer string
This issue didn't appear previously because older installs were pinned to litellm==1.67.0 (strict SemVer compliant).

Since this dependency (along with crewai) is only required for certain agents (e.g. scoring, streamlit, etc), I've moved these to an requirements-extra.txt file and updated the readme.

### DIFF
--- a/requirements-extra.txt
+++ b/requirements-extra.txt
@@ -1,0 +1,4 @@
+# Scoring Agent support
+crewai
+litellm==1.67.0
+


maximilien (approved):
LGTM

PR #533, opened 2025-06-16
fixing issue running streamlit deploy

maximilien:

PR #535, opened 2025-06-17
bug: streamlit unable to deploy after uv merge

george-lhj:
After switching to uv and src/ layout, running streamlit run src/maestro/cli/streamlit_deploy.py ... (via subprocess inside maestro deploy ... --streamlit) was failing with:

ModuleNotFoundError: No module named 'maestro.cli'
This happened because Streamlit executes the file directly (not inside uv package context). Python couldn't resolve from maestro.cli.common ... since src/ wasn't on sys.path.

fixes #533

### DIFF
--- a/src/maestro/cli/commands.py
+++ b/src/maestro/cli/commands.py
@@ -343,7 +343,7 @@ class DeployCmd(Command):
     
     def __deploy_agents_workflow_streamlit(self):
         try:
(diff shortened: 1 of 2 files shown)


maximilien:
Interesting you did not have to move package... nice. Let me try this and report here.

maximilien:
OK @george-lhj this tested fine and since small let's merge this and discard mine.a

Well done.

PR #536, opened 2025-06-17
chore: split cbom demo

george-lhj:
per meeting request to align w slides, split the CBOM demo into its original workflow and its remediation portion


maximilien (approved):
LGTM

Issue #534, opened 2025-06-16
Enhance portfolio workfkow


akihikokuroda:
The current workflow has 2 code agents.  

1. take the historical date of the given stock symbols
2. execute the Portfolio Optimizer with all default options

There are some possible enhancement:
1. take bond symbols in addition stock
2. take company names as well as stock symbols
3. allow user to provide qubo settings especially risk_aversion, restriction_coeff and transaction_fee
4. format output into easier to understand

george-lhj:
Currently, how are we getting the data of a given stock symbol? Are we using someting like yf api? Or something else?

akihikokuroda:
@george-lhj Please take any above work or something else. if you are interested in.  Thanks!

george-lhj:
I'll start with 2, then try to do 1 for the enhancements

akihikokuroda:
3 has not done but I close this.  New issues will be created if the workflow gets any feed back.

PR #537, opened 2025-06-17
demo: portfolio update

george-lhj:
in the process of implementing #534 

Got an agent working that's able to take in a list of stock or bonds names or ticker symbols and output its ticker symbols. 

I am wondering what do you think the output should be? Should it be a list of the tickers? Or json, dictionary, etc? Im not too sure.


george-lhj (commented):
if we wanted to be sequential i can make it a list, so then it can just easily pass onto the next agent that you wrote. 

from this demo we can see why we need to be able for an agent to access prevous responses, because lets say we ahve our security ticker agent, and also another agent which is determining current interest rates (or some other metrics that are needed), we need to be able to allow the portfolio agent to access them both. Even if we made another intermediate agent that gathers all teh parameters, still need to be able to access mulitiple outputs of agents.

akihikokuroda:
So far looks good.  Thanks.  The quiskit function needs data like this:
```
{
    "8801.T": {
        "2023-01-01": 2374.0,
        "2023-01-02": 2374.0,
        "2023-01-03": 2374.0,
        "2023-01-04": 2356.5,
        ...
    },
    "AAPL": {
        "2023-01-01": 145.2,
        "2023-01-02": 146.5,
        "2023-01-03": 147.3,
        "2023-01-04": 148.1,
        ...
    },
    ...
}
```
So the next agent retrieves the historical data for each symbol and format them.  The data of the date must be same for all symbols.  It may need some work when the source of data are not same.

maximilien:
Is this still supposed to be DRAFT or ready to review?

george-lhj:
@akihikokuroda  this is ready for review, see updated PR description

Issue #506, opened 2025-06-05
enhancement: maestro binary


psschwei:
Currently, the only way to run maestro is to clone the repo and then execute from the root directory. As a user, it would be nice to put maestro on my $PATH and run from the directory where I'm working.

maximilien:
Don't fight the pythonic way of things. Embrace it.

Let's do brew and pip. Then we can explore alternatives when these work.

Thanks

ajbozarth:
After some research I've come to the following:

- We already have the ability to create a wheel file via poetry build that can be pip (or pipx) installed, this would also be the artifact we would publish to pypi, this is why we can run `pip install git+https://github.com/AI4quantum/maestro`
- The setup-maestro script is no longer need, simplifying the steps to just build and run

My plan moving ford then is to write a release script with the following features:

- the workflow will trigger when a `v*` tag is pushed
- it will create a GitHub release based on the tag and will include:
    - the files in the `dist` dir as generated by a `poetry build`
    - named after the tag name
    - [auto generated release notes](https://docs.github.com/en/repositories/releasing-projects-on-github/automatically-generated-release-notes)
- publish to PyPi (once we have pypi setup, this will initially be commented out)

Open discussion issues:

- @maximilien How should we handle publishing to PyPi? There [already exists a maestro on pypi](https://pypi.org/project/maestro/) so we'd need to figure out a name and I'm unsure of what the process is for getting that approved
- Should we add automation for creating the new version tags that trigger the release? Cutting each tag would include incrementing the version in the `pyproject.toml`
- What follow up items or conflicts will need to be addressed with @maximilien plan to fetch from poetry to uv

I'm currently working on the workflow file and have a few other updates locally for build and docs, but I believe they will be unnecessary once we move to uv.  I'll try to have a draft PR open by EOD for reference.

PR #538, opened 2025-06-18
demo: add result plot step for portfolio

akihikokuroda:
The plotting only work for `run` for now.

### DIFF
--- a/demos/workflows/portfolio/README.md
+++ b/demos/workflows/portfolio/README.md
@@ -38,6 +38,8 @@ This demo shows how to use Maestro to create a portfolio management system.
 ```bash
 cd maestro
 uv pip install -e .
+uv pip install yfinace
+uv pip install qiskit-ibm-catalog
 cd -
 ```
 
--- a/demos/workflows/portfolio/agent.yaml
+++ b/demos/workflows/portfolio/agent.yaml
@@ -1,4 +1,4 @@
-apiVersion: maestro/v1alpha1
+apiVOAersion: maestro/v1alpha1
 kind: Agent
 metadata:
   name: portfolio
@@ -11,8 +11,9 @@ spec:
   instructions: call portfolio optimizer with data
   code: |
     import pandas as pd
+    import json
     from qiskit_ibm_catalog import QiskitFunctionsCatalog
-    
+
     class mock_result:
       def result(self):
         result = {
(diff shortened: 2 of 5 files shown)


akihikokuroda:
part of #534

demos/workflows/portfolio/agent.yaml
```diff
@@ -1,4 +1,4 @@
-apiVersion: maestro/v1alpha1
+apiVOAersion: maestro/v1alpha1
```
psschwei, line 1: typo?

pyproject.toml
```diff
@@ -26,7 +26,9 @@ dependencies = [
     "slack_sdk>=3.35.0",
     "nest-asyncio>=1.6.0",
     "pydantic-ai[logfire]>=0.1.8",
-    "opik>=1.7.22"
+    "opik>=1.7.22",
+    "yfinance>=0.2.63",
+    "qiskit-ibm-catalog>=0.7.0",
```
psschwei, line 31: would it be worth adding dependencies for the demos in their own [dependency group](https://docs.astral.sh/uv/concepts/projects/dependencies/#dependency-groups)?

george-lhj (commented):
the agent file name should be could `agents.yaml` to match the other workflows, other after the changes good to merge

maximilien:
LGTM once comments addressed

akihikokuroda:
All comments were addressed.

demos/workflows/portfolio/README.md
```diff
@@ -38,6 +38,8 @@ This demo shows how to use Maestro to create a portfolio management system.
 ```bash
 cd maestro
 uv pip install -e .
+uv pip install yfinace
+uv pip install qiskit-ibm-catalog
```
psschwei, line 42: I think this is fine for now, but longer-term adding something like this to the pyproject.toml might be better

```toml
[project.optional-dependencies]
demo_portfolio = [
  "yfiance",
  "qiskit-ibm-catalog"
]
```

then you could install with something like `uv pip install .[demo_portfolio]`

psschwei (commented):
LGTM

demos/workflows/portfolio/README.md
maximilien: What about separate `requirements-demos.txt` file?

What's the best practice?

psschwei: if you're using pyproject.toml, I think best practice is to put all dependencies in there rather than having an additional requirements.txt file -- that will let you also reconcile the dep versions between main and demo(s) too, which you wouldn't get with multiple files

I can look into making a PR for that tomorrow (unless someone else really wants to tackle it)

ajbozarth: > if you're using pyproject.toml, I think best practice is to put all dependencies in there rather than having an additional requirements.txt file -- that will let you also reconcile the dep versions between main and demo(s) too, which you wouldn't get with multiple files

I'm actually actively working on this change rn as part of #506

akihikokuroda: I'll leave this to Alex handling in #506

maximilien (approved):
/lgtm

PR #539, opened 2025-06-18
chore: switching licenses to SPDX-syle

maximilien:
Convinced this is a better approach if only because:

1. shorter
2. reduces possibility that a file has wrong license text

I know @planetf1 had started doing that... this does for all at once. Thanks for suggestion Nigel.


maximilien:
Got the 👍 from @akihikokuroda so merging to move fast.

Issue #532, opened 2025-06-16
bug: remaining issues with uv switch


maximilien:
Let's use this issue for any remaining issues:

- [x] Operator Tests CI
- [x] Streamlit
- [ ] [other]

CC: @akihikokuroda @george-lhj

maximilien:
I believe @akihikokuroda fixed the first issue with PR #531

ajbozarth:
I had this as a subtask in #541 but its better to add here:

- [ ] Look into a better way to call tests (commands like uv run lint and uv run tests are not supported out of the box by uv afaik)

ajbozarth:
I believe I figured out an easy way to deal with the test/lint, I'll open a PR once I've gotten it sorted

ajbozarth:
I opened https://github.com/AI4quantum/maestro/issues/562 which could be considered a subtask of this

PR #541, opened 2025-06-19
feature: support pip installing maestro

### LINKED ISSUE #506: enhancement: maestro binary

ajbozarth:
Closes #506 

- Adds a two GitHub Actions:
    1. `maestro_release.yaml` to create GitHub release when a `v*` tag is created. These releases can then be pip installed.
    2. `maestro_test-workflow.yaml` which pip installs maestro and runs a new `simple` demo

- Updates the pyproject.toml to handle installing dependencies and removes (now) redundant build files

    How to install with the new steps:

    - `uv pip install -e .` will install a local dev install (note the `uv pip` when doing dev installs)
    - `uv pip install -e .[crewai]` will also install the optional crewai dependencies 
    - `pip install git+https://github.com/AI4quantum/maestro.git@v0.1.0` will install a release (specifically v0.1.0 here)
    - `pip install "maestro[crewai] @ git+https://github.com/AI4quantum/maestro.git@v0.1.0"` will do the same but with the crewai dependencies
    - You can also pip install a release wheel file by referencing the GitHub release artifact url or downloading it locally

- Updates the docs and GitHub workflows with the new install commands

- Also include some bug fixes found when running maestro via pip install

To Do:
- [x] Test that installing via the wheel file can still run all demos
- [x] Further clean up and update docs
- [x] Update other workflows with new install steps and clean up unnecessary requirements files

Followup:
- Look into a better way to call tests (commands like `uv run lint` and `uv run tests` are not supported out of the box by uv afaik) (moved to #532)
- Improve demo docs, this can be done when we move demos into their own repo
    - When we create the separate demo repo we will need to include the schemas dir as well to support the `maestro validate` steps in some demos
    - the meta-agent demo docs need a lot of improvement, they are not very clear about what is being done and how to do it


psschwei (commented):
Doing pip install via the link to the wheel on your fork:

First time I got `ERROR: Package 'maestro' requires a different Python: 3.13.3 not in '<3.13,>=3.11'`

Once I dropped to an appropriate Python version, it installed fine and I was able to successfully run the `portfolio` demo.

george-lhj (commented):
nice, the demos are able to run except for the scoring agents because this specific command `pip install -e .[scoring]` doesn't work for me. 

zsh: no matches found: .[scoring]

ajbozarth:
found the correct way to pip install a wheel file via url with extras:
```
pip install "maestro[scoring] @ https://github.com/ajbozarth/maestro/releases/download/v0.1.1/maestro-0.1.0-py3-none-any.whl"
```

I'll update the top description with all the changes and I'll update the docs with my next edit

ajbozarth:
> Would you add the way to install the release from github (until it goes into pypi) in the README.md

That is the plan, it's part of the `Further clean up and update docs` To Do item

maximilien (changes requested):
As per discussion yesterday please ensure you have a GH action that verifies this feature on a fresh container mimicking what a pure maestro user would do. So something like:

1. Create new directory 
2. Install Python and pip
3. Pip install maestro and dependencies 
4. Write new agents.yaml and workflow.yaml
5. Execute workflow via CLI and deploy it

Once we have this ^^^ or similar then I would say this is mergeable.

Ideally steps above would be executed in fresh clean container and everything added to it.

Thanks.

ajbozarth:
I've updated the docs and the existing CI (still trying to get the test CI to pass) and removed the old build files.

I wanted to remove the `maestro` executable, but hit a couple snags, it's still used by the `tools/check-mermaid.sh`, `tools/check-schemas.sh`, and `tools/update_readmes.sh` scripts, which are used in CI.

Currently if you `pip install` maestro you can just use the `maestro` command, but if you dev install with `uv pip install` then the command is `uv run maestro` instead. The executable bypasses this.

My next focus will be finishing testing all the examples in a clean env with pip installed maestro

ajbozarth:
Record of which demos I've tested (I started over testing each after my last commit):

- [ ] activity-planner-crewai.ai
- [ ] activity-planner.ai
- [ ] cbom.ai
- [ ] ibm-summary.ai
- [ ] mcp-gh-tools.ai
- [ ] meta-agents
- [ ] openai-mcp.ai
- [ ] portfolio
- [ ] qaoa
- [ ] simple
- [ ] summary.ai
- [ ] weather-checker.ai

This is just for my own tracking and to give insight on how final testing is going

ajbozarth:
Just pushed an update that fixes the `maestro deploy` issue, it was a file naming conflict, solved by renaming the file. I also hard-coded the references to the image file since it isn't found when running as a package.

With this I've now tested summary.ai and weather-checker.ai and cleaned up their docs a bit (removed wrong parts and cleaned up the steps). I'll probably do a bit of doc clean up as I continue, but I'll be calling it a day here.

.github/release.yml
maximilien, line 1: I assume these different sections will be auto-populated? Should be easy since we are pretty good (with me as a cop ;) ensure PRs and issues are labeled.

.github/workflows/maestro_release.yaml
```diff
@@ -0,0 +1,32 @@
...
+      - name: Build Maestro
+        shell: bash
+        run: uv build
+      - name: GitHub Release
+        uses: softprops/action-gh-release@v2
+        if: github.ref_type == 'tag'
+        with:
+          files: dist/*
+          generate_release_notes: true
+#      - uses: actions/upload-artifact@v4
```
maximilien, line 28: Why these commented lines? Get rid of them. It's in git log if you need to ever go back.

CONTRIBUTING.md
```diff
@@ -43,16 +40,16 @@ git push origin feature/your-feature-name
 
 ## Code Style
 
-We use [black](https://github.com/psf/black) for code formatting and [flake8](https://flake8.pycqa.org/) for linting. To ensure your code meets our style guidelines:
+We use [black](https://github.com/psf/black) for code formatting. To ensure your code meets our style guidelines:
```
maximilien, line 43: We may want to look into ruff which is also created by the same folks who created uv

maximilien: Unsure if the overlap is perfect...

README.md
```diff
@@ -4,57 +4,53 @@ Maestro is a tool for managing and running AI agents and workflows.
...
-```bash
-uv venv
-source .venv/bin/activate
-uv pip install -r requirements.txt
-uv pip install -e .
+pip install git+https://github.com/AI4quantum/maestro.git@v0.1.0
 ```
 
-Note: If using scoring agents, additionally install:
+Note: If using crewai agents, install:
```
maximilien, line 11: Actually it's the scoring agent... check with @george-lhj

maximilien (changes requested):
See comments. Also please ensure either or both @akihikokuroda and @george-lhj gives their OK.

@AngeloDanducci could also be a good person to verify the demos parts.

Thanks for this @ajbozarth

CONTRIBUTING.md
ajbozarth: I originally had planned to address this in this PR but decided instead to list it as a follow up item in #532 

There is definitely improvements to how we lint/test, this cleanup is because we currently don't include flake for linting.

README.md
ajbozarth: I had originally called the extra `scoring` as seen in my demo Monday, but when reading the code further the warning was coming from the `CrewAIAgent` so I changed it.

@george-lhj which is correct? Either way I think naming the extra package `crewai` is better since that's what is actually installs

ajbozarth:
For those testing you can pip install the version of maestro in this PR with:
```
pip install git+https://github.com/ajbozarth/maestro.git@v0.1.0
```
That's a release I cut on my fork base on the current version on this PR

ajbozarth:
> Should we document that the user needs to set the BEE_AI/API_KEY themselves? Also it seemed like they weren't being loaded for me, I had to manuallly set them to get the demos to work

@george-lhj from my testing if you have the .env file located in the directory you run maestro from it should work, but you are correct I did not update any docs to clarify that. If can confirm that works then I will update the relevant docs

ajbozarth:
> if you have the .env file located in the directory you run maestro from it should work

Also on this I think it makes sense to move the example env to a common directory and update all the doc references accordingly. (and makes sure the demos that need the env vars in the example env are clear about it)

george-lhj:
for me this is working from demos side, I think others should also review since big PR?

maximilien:
you need to rebase @ajbozarth should be easy

ajbozarth:
> you need to rebase 

done

maximilien:
@ajbozarth since you want to cut a release after this let's make sure we merge chore #555 which fixes various small items on markdown docs files. This will ensure the release includes these changes.

ajbozarth:
> let's make sure we merge chore https://github.com/AI4quantum/maestro/pull/555 which fixes various small items on markdown docs files. This will ensure the release includes these changes.

That should be ok, but that PR will need a lot of rebasing once this is merged since this PR changes a majority of the same lines edited in that PR (eg the ones around build commands). Similarly even more would conflict with the change to ruff I plan to open once this is merged (https://github.com/ajbozarth/maestro/compare/release...ajbozarth:maestro:ruff)

ajbozarth:
I believe I've cleaned up the doc issues from the merge. This should be good for final review and merge. 

Once it's merged I'll be opening the follow up PR to switch to using ruff, which is non-blocking to a v0.1.0 release

src/maestro/cli/streamlit_deploy.py
```diff
@@ -45,7 +45,7 @@ def deploy_agents_workflow_streamlit(agents_file, workflow_file):
         st.session_state.messages = [
             {"role": "assistant", "content": "Welcome to Maestro workflow"}]
 
-    st.image("images/maestro.png", width=200)
+    st.image("https://raw.githubusercontent.com/AI4quantum/maestro/main/images/maestro.png", width=200)
```
akihikokuroda, line 48: Is it possible to include the image file in the package?  I may not good to access to github when the maestro command runs.

src/maestro/cli/streamlit_meta_agents_deploy.py
```diff
@@ -51,7 +51,7 @@ def load_generated():
         ]
 
     # Page header
-    st.image("images/maestro.png", width=200)
+    st.image("https://raw.githubusercontent.com/AI4quantum/maestro/main/images/maestro.png", width=200)
     st.title("Maestro Meta-Agents workflows")
```
akihikokuroda, line 55: same as above.

README.md
```diff
@@ -4,57 +4,53 @@ Maestro is a tool for managing and running AI agents and workflows.
...
-cd maestro
-```
-
-2. Install dependencies:
-```bash
-uv venv
-source .venv/bin/activate
-uv pip install -r requirements.txt
-uv pip install -e .
+pip install git+https://github.com/AI4quantum/maestro.git@v0.1.0
```
akihikokuroda, line 8: Are we updating the file for each release? Are there any other files that we have to update for the new release?  It may be better to putting something like "release number" instead of the real release number.  WDYT?

ajbozarth: I had thought about it and finding there wasn't a way to dynamically update the command string with the latest version, and came to two possible conclusions:

1. We use the latest version like I did here and just remember to update it alongside the pyproject.toml each release
2. We replace `v0.1.0` with `<release>`, which forces the user to add the version afterr pasting the command.

I went with option 1 because it has better usability for users, and eventually we intend to just release this on pypi and update this command to `pip install maestro` (with whichever name we replace maestro with), so we would only have to deal with this minor release update until that point. Plus we already have to manually update the pyproject.toml prior to cutting a release tag anyway.

I'm not stuck on this though and am willing to switch to option 2 if you have a string preference

> Are there any other files that we have to update for the new release?

Once we move the demos into their own repo we would only have to update this line in the README and the `pyproject.toml`, but this pip install command is also found in a handful of the demo READMEs ad well

src/maestro/cli/streamlit_deploy.py
ajbozarth: I'll take a look into this, this was more of a fix for when running the demos from a separate directory than the git repo, but you are correct that it's better to include them in the python package and use them directly.

README.md
akihikokuroda: OK.  Sounds reasonable to me.  Would you add a section a "release process" in the CONTRUBUTING.md and include the list of files must be updated for each release?
Thanks!

.github/workflows/maestro_release.yaml
```diff
@@ -0,0 +1,27 @@
...
+        uses: astral-sh/setup-uv@v6
+      - name: Build Maestro
+        shell: bash
+        run: uv build
+      - name: GitHub Release
+        uses: softprops/action-gh-release@v2
+        if: github.ref_type == 'tag'
+        with:
+          files: dist/*
+          generate_release_notes: true
```
akihikokuroda, line 27: Would you add a sniff test step?  It is probably enough to install the package and run a existing test.

README.md
ajbozarth: > Would you add a section a "release process" in the CONTRUBUTING.md

done

.github/workflows/maestro_release.yaml
ajbozarth: done, I added a test job and made it a prerequisite for the release job, I also confirmed it works as expected on my fork

akihikokuroda: I might not explain correctly what I want.  Can the package just build be tested?

ajbozarth: > Can the package just build be tested?

I was eventually able to add a test like in `maestro_test-workflow.yaml`, but our current pytests would require quite a bit of editing to support running against a distribution install rather than a dev install as they were intended.

ajbozarth:
> Would you make a test release to make sure the CI work?
 
see my fork: https://github.com/ajbozarth/maestro

akihikokuroda:
The test step executed in the same job as the build step.  It is not clear if the test executing the code installed from the package or the repository source.  Would you make a follow up PR for separating them?

ajbozarth:
> Would you make a follow up PR for separating them?

I just did it here (should've done it from the start)

maximilien (approved):
LGTM since @george-lhj and @akihikokuroda gave their OK

PR #542, opened 2025-06-19
gitignore files created by portfolio demo

psschwei:
If you run the demo, there's a few CSV files that get created. This is a small fix to not accidentally commit those files to Github.


akihikokuroda (approved):
Thanks!

PR #543, opened 2025-06-19
feature: add some qiskit MCP tools

akihikokuroda:
This PR has some qiskit MCP tools.  This probably is the temporary location for these tools.
All tools have mock code for the qiskit function call.

`mcp/mcptools/qiskitmcp.py` has following tools:
```
tools run_iskay_optimizer
tools run_qctrl_optimizer
tools run_qunasys_quri_chemistry
tools get_quri_chemistry_job_status
tools get_quri_chemistry_job_result
tools get_optimizer_job_status
tools get_optimizer_result
```

`mcp/examples/mcp_qiskit.py` is an example agent that loads and uses all these tools.  It is based on the BeeAgent example.  It can be run by

```
python mcp/examples/mcp_qiskit.py
```

The following is an example run:
```
User 👤 : run iskay optimizer with this as the objective function objective_func = {"()": 1,"(0,)": 1.5,"(1,)": 2,"(2,)": 1.3,"(0, 3)": 2.5,"(1, 4)": 3.5,"(0, 1, 2)": 4,}
Agent(final_answer) 🤖 :  The Iskay optimizer job is DONE.
User 👤 : get the Iskay optimizer job status
Agent(final_answer) 🤖 :  The Iskay optimizer job is DONE.
User 👤 : retrieve the result o Iskay optimizer job
Agent(final_answer) 🤖 :  The solution to the Iskay optimizer job is {"0": -1, "1": -1, "2": -1, "3": 1, "4": 1} with a cost of -13.8.
```


maximilien:
> I notice that there's a lot of similarity in some of the functions... what do you think about doing something a bit more generic: like a `run_function(function_name)` tool where the actual function can be parameterized, a `get_function_results(job_id)`, etc... that would also let the MCP server manage the functions a bit more dynamically ?

On one hand I appreciate the dry nature of @psschwei request on the other I prefer this approach. Mainly because for each tool / function the params are different and by defining different functions with comment and args definition this might make the tool easier to call from the caller agent perfective.

akihikokuroda:
@psschwei the decorator `@mcp.tool()` does the magic.  The it takes the function protptype and comments of the function and generates name, descriptions and input schema, etc.  So putting them into a generic function confuses the decorator and mess the tool definitions.

mcp/examples/mcp_qiskit.py
```diff
@@ -0,0 +1,173 @@
...
+from beeai_framework.errors import FrameworkError
+from beeai_framework.logger import Logger
+from beeai_framework.memory import TokenMemory
+from beeai_framework.tools import AnyTool
+from beeai_framework.tools.mcp import MCPTool
+from beeai_framework.tools.code import SandboxTool
+
+# Example prompt
+#
+# solve following cost function and retrive the result {"()": 1,"(0,)": 1.5,"(1,)": 2,"(2,)": 1.3,"(0, 3)": 2.5,"(1, 4)": 3.5,"(0, 1, 2)": 4}
```
maximilien, line 30: Does this example belong here or a README.md?

mcp/examples/mcp_qiskit.py
maximilien, line 1: We should create another example with different agent framework or best using `maestro` :)

Let's discuss this.

mcp/mcptools/qiskit_mcp.py
```diff
@@ -0,0 +1,230 @@
+from typing import Any
+import httpx
+from mcp.server.fastmcp import FastMCP
+#from qiskit_ibm_catalog import QiskitFunctionsCatalog
```
maximilien, line 4: remove

mcp/mcptools/qiskit_mcp.py
```diff
@@ -0,0 +1,230 @@
...
+        options: options to handle hardware submission. format is dictionary.  shots: nunber of iteration, num_iterations: total number of Bias Field iterations, use_session: where to run within a session    
+    Ouput:
+        job_id: job id of the optimization job
+    """
+    # catalog = QiskitFunctionsCatalog()
+ 
+    # Access Function
+    # optimizer = catalog.load("kipu-quantum/iskay-quantum-optimizer")
+
+    arguments = {
```
maximilien, line 30: I don't understand this implementation. Creates a local arguments object which will be out of scope once call is return?

mcp/mcptools/qiskit_mcp.py
```diff
@@ -0,0 +1,230 @@
...
+        "problem": problem,
+        "problem_type": problem_type,
+        "instance": instance,
+        "backend_name": backend_name,  # such as "ibm_fez"
+        "options": options,
+    }
+ 
+    # job = optimizer.run(**arguments)
```
maximilien, line 38: I see so this is commented. Maybe because you cannot run now. Could we setup a mock and enable real when we can with some flag?

mcp/mcptools/qiskit_mcp.py
maximilien, line 1: I see so code is commented since you still working on access.

What I would suggest is to set up mocks and we can have a global flag that turns mock on or off. This will be useful for testing and also for demos. WDYT?

maximilien (changes requested):
See comments.

mcp/mcptools/qiskit_mcp.py
akihikokuroda: This is necessary for the real qiskit function call. I'll think about how to mock the qiskit function and about this line.

mcp/mcptools/qiskit_mcp.py
akihikokuroda: This is converting the input to the MCP function to the input for the qiskit function.  I'm not sure if I understand your question.  Thanks!

maximilien (approved):
LGTM unless there are still comments to address.

akihikokuroda:
Yes, I'm still working on the mock.

akihikokuroda:
I merge this.  Any necessary changes will be in follow on PRs.

PR #547, opened 2025-06-24
fix broken agent command in readme

psschwei:
`maestro run <agent_path>` is an invalid command
also we're using apache not mit license


george-lhj (approved):
good catch

ajbozarth (approved):

PR #549, opened 2025-06-24
enhancement: adding local workflow logging

george-lhj:
This PR introduces a comprehensive local logging system for Maestro workflows, backed by a filesystem-based strategy (e.g., ~/.maestro/logs). It also establishes a future-ready foundation for deeper logging hooks via decorators and structured step tracking.

1. File-Based Workflow Logging
Each maestro run execution generates a dedicated log file:
~/.maestro/logs/maestro_run_{workflow_id}.log

Workflow Logs include:
```
workflow_id: unique identifier
timestamp: UTC execution time
workflow_name: from workflow YAML
prompt: initial input
output: final result (excluding slack_agent, scoring_agent)
models_used: names or fallback (code:{agent_name})
status: success or error
```

2. Agent-Level Step Logging
Each step execution logs detailed info:
```
Step index
Agent name & model
Input/response text
Tool used (if applicable)
Duration (ms)
```

3. Final output selection skips known non-user-visible agents (e.g., slack_agent, scoring_agent) to better reflect actual workflow results.

4. Default Log Directory
Logs are stored at: ~/.maestro/logs/

How to Test
Run any workflow:

```
maestro run ./workflow.yaml

cat ~/.maestro/logs/maestro_run_<id>.log
```

Example log entry:

```
Timestamp     : 2025-06-24T21:55:32.079065+00:00
Workflow Name : cbom.ai
Prompt        : Use the git fetcher tool to get...
Output        : {...}
Models Used   : ["code:gh fetch", "granite3.2:latest", ..., "qwen3:latest"]
Status        : success
--- Agent Response ---
Step Index    : 0
Agent Name    : gh fetch
Model         : code
Input         : Use the git fetcher...
Response      : {...}
Duration (ms) : 231
```

Separated logging logic from internal run logic, now it is called as a decorator


src/maestro/cli/commands.py
```diff
@@ -300,27 +302,87 @@ def name(self):
...
+            else:
+                agents_yaml = None
+                Console.warn("⚠️ No agents.yaml path provided or found — skipping custom_agent label handling.")
         if self.prompt():
             prompt = self.__read_prompt()
             workflow_yaml[0]['spec']['template']['prompt'] = prompt
-
+        workflow_id = generate_workflow_id()
         try:
-            self.__run_agents_workflow(agents_yaml, workflow_yaml)
```
maximilien, line 317: Wondering if it would make sense to
Move all the code dealing with DB into private method? That way run is clean amwirh try catch and any pre and post processing?

src/maestro/db_logger.py
maximilien, line 1: This screams for an DbLogger class instead of a series of functions? Would allow these loggers to be used multiple times if needed or unsure a singleton? Ping me if you want more details on that idea.

maximilien (changes requested):
Two important comments / request for changes

src/maestro/cli/commands.py
george-lhj: yup done

src/maestro/db_logger.py
george-lhj: makes sense, done

psschwei:
I left this comment in the issue, but will reproduce here:

Thinking long-term, any solution that runs at scale is going to be on Kubernetes, which has its own apparatus for collecting logs (capturing anything written to stdout/stderr).

For local logging, what about starting with something simple, like setting up a ~/.maestro directory and writing logs from each run to a new file there?

psschwei:
It's also worth noting that maestro as of now is a CLI. It's not clear how a database fits in with a CLI. 
If maestro is evolving into something else, that's fine. But I haven't seen any discussion of that as of yet.

george-lhj:
@psschwei are you thinking of something like this more?

maximilien (changes requested):
Good progress and surely you know what's next... tests :)
You added a new class and file please add a test. Does not have to be exhaustive but should test corner cases etc.
Thanks @george-lhj

george-lhj:
done @maximilien

psschwei:
> psschwei are you thinking of something like this more?

yeah, something like that. my main thought was to keep things as simple as possible at the beginning. but that said, if you have reasons for wanting / needing to use a DB, then you should use a DB.

src/maestro/file_logger.py
```diff
@@ -0,0 +1,37 @@
...
+
+class FileLogger:
+    def __init__(self, log_dir=None):
+        self.log_dir = Path(log_dir) if log_dir else Path.cwd()
+        self.log_dir.mkdir(parents=True, exist_ok=True)
+
+    def generate_workflow_id(self):
+        return uuid.uuid4().hex
+
+    def log_workflow_run(self, workflow_id, workflow_name, prompt, output, models_used, status):
```
psschwei, line 13: question about the log formatting: will you be able to read these files into something like pandas or a database without too much work if you want to run analysis on them?

src/maestro/file_logger.py
```diff
@@ -1,10 +1,11 @@
 import uuid
 from datetime import datetime, UTC
 from pathlib import Path
+import os
 
 home_path = Path.home()
-if home_path == Path("/"):
-    DEFAULT_LOG_DIR = Path("/app/logs")
+if not os.access(home_path, os.W_OK):
+    DEFAULT_LOG_DIR = Path("/tmp/maestro/logs")
```
psschwei, line 8: these will get deleted if the user shuts down their laptop... maybe somewhere like `./logs` for persistence?

george-lhj:
updated to #560

PR #553, opened 2025-06-26
fix: remove crewai warning message

ajbozarth:
This remove the warning message about crewai not being installed from showing on every maestro command.

We already had an error handler for the case where the `CrewAIAgent` is used without crewai installed, so I moved the install command message into that error message.


ajbozarth:
There has been discussion around removing this so here's a potential fix

PR #561, opened 2025-06-30
test: switch to using ruff for format and lint

ajbozarth:
This adds support for code formatting and littering using ruff and included the relevant doc updates.

It also includes a pre-commit config that is not enabled by default.

> [!IMPORTANT]  
> This PR does not enforce any formatting or linting, it only adds the dev dependencies, configs, and documentation

I have also done an initial run of the formatter and linter and committed it in a separate branch: https://github.com/AI4quantum/maestro/commit/99ebdae410ea418c7a98dd1bb56420d6df3a9bd1

If desired I can include that commit in this PR, otherwise it will be opened as a followup draft PR for later

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,9 @@
+repos:
+- repo: https://github.com/astral-sh/ruff-pre-commit
+  # Ruff version.
+  rev: v0.12.0
+  hooks:
+    # Run the linter.
+    - id: ruff-check
+    # Run the formatter.
+    - id: ruff-format
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -23,6 +23,11 @@ source .venv/bin/activate  # On Unix/macOS
 uv pip install -e .
 ```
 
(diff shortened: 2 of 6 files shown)


george-lhj (commented):
Unsure if everyone is aware/approved to switch from black to ruff (or maybe I just missed when this was discussed), but I've finished reviewing the functionality works

I'll wait for @psschwei or @maximilien to comment about the lint switch, since personally I don't really have an opinion on this

psschwei (commented):
I do not have a strong opinion about linting tools, as long as we pick one and it works :smile: That said, ruff seems to have a lot of traction these days, so I'd be +1 for switching over.

ajbozarth:
>Unsure if everyone is aware/approved to switch from black to ruff

There was no discussion about switching the formatter from black to ruff, the discussion was about which linter we should use when adding lint support.

When choosing a linter, ruff was the first choice because it's the ruff team that created uv. Then, since we want to use ruff for linting it makes sense to also use it for formatting since it does both, which is the reason for switch away from black.

PR #564, opened 2025-07-01
test: fix buggy streamlit tests

### LINKED ISSUE #562: bug: test_validate_create_run passing when it should be failing
The test `test_validate_create_run` currently prints out the following error when run:

```
Error: Invalid value: File does not exist: /Users/ajbozarth/workspace/quantum/ai/maestro/src/maestro/cli/cli/streamlit_meta_agents_deploy.py
```

Then it opens a streamlit instance despite it supposed to be a dry run. This is due to two separate but intertwined bugs:

1. the streamlit command that is run by the `deploy` and `meta-agents` commands is not respecting the dry-run flag and is opening streamlit instances when running the test. We only see one such instance currently due to above error.
2. The streamlit command in `meta-agents` has a malformed path, but once that path is corrected the test then starts failing due to various other issues

ajbozarth:
Fixes #562 

Address some silently failing tests:
- the `--kubernetes` flag was being ignored causing the `test_deploy__dry_run_kubernetes` test to run a not mocked local deploy and start streamlit
- `test_meta_agents` was silently failing due to some missed uv updates to `streamlit_meta_agents_deploy.py`, ionce fixed it now spins up a streamlit instance, which I added a cleanup step for.

### DIFF
--- a/src/maestro/cli/commands.py
+++ b/src/maestro/cli/commands.py
@@ -333,7 +333,7 @@ class DeployCmd(Command):
     def __deploy_agents_workflow_streamlit(self):
         try:
             sys.argv = ["uv", "run", "streamlit", "run", "--ui.hideTopBar", "True", "--client.toolbarMode", "minimal", f"{os.path.dirname(__file__)}/streamlit_deploy.py", self.AGENTS_FILE(), self.WORKFLOW_FILE()]
-            process = subprocess.Popen(sys.argv)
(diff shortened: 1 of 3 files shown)


george-lhj (approved):
this fixes the streamlit launching, thanks! lgtm

maximilien (approved):
Nice. Thanks @ajbozarth

PR #565, opened 2025-07-01
refactor: apply ruff format and lint to code

ajbozarth:
This is a follow up to #561 that applies the ruff format and lint.

This PR contains all the changes generated by, or addressing the output of:
```bash
uv run ruff format
uv run ruff check --fix
```

Per prior discussion, this is in draft until we decide we want to apply the changes, since they touch nearly every file.

In addition it's worth noting that this PR still does not include enforcing lint, just applying it to the current codebase

### DIFF
--- a/src/maestro/cli/commands.py
+++ b/src/maestro/cli/commands.py
@@ -21,6 +21,8 @@ from jsonschema.exceptions import ValidationError, SchemaError
 from maestro.deploy import Deploy
 from maestro.workflow import Workflow, create_agents
 from maestro.cli.common import Console, parse_yaml
+from maestro.file_logger import FileLogger
+from datetime import datetime, UTC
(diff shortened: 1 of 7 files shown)


psschwei:
I don't see any harm in merging this (assuming all tests pass), as it's just formatting. Heck, it seems like you can even configure [many editors](https://docs.astral.sh/ruff/editors/setup/) to use ruff on save (similar to gopls for golang).

ajbozarth:
> I don't see any harm in merging this

I'll move it out of draft so it can be merge if desired

> you can even configure [many editors](https://docs.astral.sh/ruff/editors/setup/) to use ruff on save

I actually had to deactivate this in my IDE while working on maestro because it kept editing every file I touched

george-lhj:
thanks for finding it, I fixed it and pushed a pr

PR #566, opened 2025-07-01
feature: add wow support

### LINKED ISSUE #513: feature: workflows of workflows (WoW)
In many use cases the resulting workflow will be large with many agents. Like for all good software engineering the solution will be to decompose into many smaller workflows or workflows of workflows.

Besides easier debugging, testing, and development, such WoW will enable the ability to have logic between the workflows as they execute. For instance, to allow retries, deal with error, etc.

CC: @george-lhj @akihikokuroda

akihikokuroda:
fix #513

This assumes that the workflow is served remotely.

### DIFF
--- a/schemas/workflow_schema.json
+++ b/schemas/workflow_schema.json
@@ -77,6 +77,22 @@
                 "type": "string"
               }
             },
+            "workflows": {
+              "type": "array",
+              "items": {
+                "type": "object",
+                "properties": {
+                  "name": {
+                    "type": "string",
+                    "description": "name of workflow"
+                  },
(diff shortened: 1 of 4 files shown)


george-lhj (commented):
oh sorry, i had to make some changes in step and workflow to support logging, so you probably have to refactor a bit

maximilien (changes requested):
Love the first attempt. We do need to figure out what happens if the workflow fails and how to propagate exception...

Also we definitely need tests. But this is a good start. Thanks @akihikokuroda

PR #567, opened 2025-07-01
bug: removed extra def

george-lhj:
forgot to delete additional lines from run command which were overwritten. 

credit: @ajbozarth for finding this, thank you!


ajbozarth (approved):
LGTM, same fix as I had put in #565

PR #571, opened 2025-07-02
bug: strip thinking output from agent response

### LINKED ISSUE #568: feature/bug: parsing output from agent
For a few of the models (like deepseek, qwen), the LLM provides output to their iterative thinking process, and then gives the actual output:

```🐝 Response from TaskInterpreter: <think>
Okay, so the user wants me to figure out how to fetch current stock prices for Apple and Microsoft and then analyze which one has performed better over the past week. Let's break this down step by step.

First, I need to get the current stock prices. For that, I can use an API like Alpha Vantage since it's a reliable source for real-time financial data. So, I'll create an agent called stock_data_retriever that fetches the latest prices for both companies.

Next, once I have the current prices, I need to look back over the past week to compare their performance. To do this, I can use Yahoo Finance API because it provides historical market data. Another agent, stock_historical_analyzer, will retrieve the weekly performance data for both stocks.

Now, with the current prices and historical data, I need to calculate which stock performed better. For that, a third agent, stock_performance_calculator, will compute the percentage change in stock price over the past week for both Apple and Microsoft. This will help determine which one had a better performance based on the increase or decrease.

Wait, but the user mentioned using two agents in the examples. Maybe I can combine the data retrieval and analysis into a single agent to keep it simple. So instead of having three separate agents, perhaps I can have stock_data_retriever collect both current and historical data, and then stock_performance_analyzer do the comparison.

Hmm, but would that be efficient? Let me think. If I retrieve all the data in one go, the analysis might require more complex calculations. Alternatively, maybe it's better to separate the data retrieval from the analysis to keep things clear. So, using three agents makes sense: two for fetching data and one for analyzing.

But since the user prefers simplicity and minimal agents unless necessary, perhaps combining into two agents would be better. Let me see: first retrieve current prices and historical data (agent1), then calculate performance (agent2). But wait, agent2 might need both datasets to perform the analysis, so maybe it's still three steps.

Alternatively, have agent1 fetch both current and 1-week history for each stock, and agent2 compute the performance comparison. That way, we only use two agents: one for data retrieval and another for analysis.

I think that makes sense. So, agent1 will handle fetching all necessary data, and agent2 will process it to determine which stock performed better.
</think>
```
And then the actual output
```
number of agents: 2
agent1: stock_data_retriever – Retrieves current stock prices for Apple and Microsoft, as well as their historical performance over the past week.
agent2: stock_performance_analyzer – Analyzes the historical data to determine which stock has performed better over the past week.
```

The problem with this is that it affects the next input to the agent. Instead of only passing the desired output, it also passes the thinking stage. 

We need to be able to parse and make sure when we pass the input, it does not have the information inside of <think> (but keep displaying it in terminal is ok)

george-lhj:
fixes #568

### DIFF
--- a/src/maestro/step.py
+++ b/src/maestro/step.py
@@ -10,6 +10,8 @@ load_dotenv()
 
 
 def strip_think_tags(text: str) -> str:
+    if not isinstance(text, str):
+        return text
     return re.sub(r"<think>.*?</think>", "", text, flags=re.DOTALL).strip()
 
 


george-lhj:
ill make a new PR

george-lhj:
closed bc includes my meta agent testing

PR #574, opened 2025-07-02
fix: allow cors for maestro serve

ajbozarth:
Updates `maestro serve` to support a configurable env var `CORS_ALLOW_ORIGINS` for CORS. This comma-delimitated var is read from the env at the startup of maestro serve, and by default is empty

### DIFF
--- a/src/maestro/agents/openai_agent.py
+++ b/src/maestro/agents/openai_agent.py
@@ -26,6 +26,9 @@ from agents.extensions.models.litellm_model import LitellmModel
 from maestro.agents.agent import Agent as MaestroAgent
 from maestro.agents.openai_mcp import setup_mcp_servers, MCPServerInstance
 
+from dotenv import load_dotenv
+
+load_dotenv()
 
 SUPPORTED_TOOL_NAME: Final[str] = "web_search"
 TOOL_REQUIRES_RESPONSES_API: Final[bool] = True
--- a/src/maestro/cli/fastapi_serve.py
+++ b/src/maestro/cli/fastapi_serve.py
@@ -210,13 +210,13 @@ class WorkflowHealthResponse(BaseModel):
     status: str
     workflow_name: str
     timestamp: str
- 
+
 class FastAPIWorkflowServer:
     """FastAPI server for serving Maestro workflow."""
-    
+
     def __init__(self, agents_file: str, workflow_file: str):
         """Initialize the FastAPI server.
-        
+
         Args:
             agents_file: Path to the agents YAML file
             workfloe_file: Path to the workflow YAML file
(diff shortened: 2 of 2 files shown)


ajbozarth:
leaving in draft for discussion, my web app POC is pointing to this branch so we don't need to merge yet

src/maestro/cli/fastapi_serve.py
```diff
@@ -58,6 +59,13 @@ def __init__(self, agents_file: str, agent_name: Optional[str] = None):
             description="HTTP API for serving Maestro agents",
             version="1.0.0",
         )
+        self.app.add_middleware(
+            CORSMiddleware,
+            allow_origins=["*"],
+            allow_credentials=True,
+            allow_methods=["*"],
+            allow_headers=["*"],
```
psschwei, line 67: I think ideally there would be config options for these in the agent spec somehow, with defaults in case they aren't specified

ajbozarth: would it be better to config them in the env instead? I'm unsure if adding serve configs to the agent is the best solution. Also which values should have configs? just `allow_origins` or all four of them?.

psschwei: We could probably just start with allow_origins... if others are needed we should be able to build off that approach (and it lets us add user knobs slowly only if there's need)

As for agent spec vs. envvars, I don't really have a strong opinion either way.

akihikokuroda:
@ajbozarth I'm not sure if they all need to be "*".

src/maestro/agents/openai_agent.py
```diff
@@ -26,6 +26,9 @@
 from maestro.agents.agent import Agent as MaestroAgent
 from maestro.agents.openai_mcp import setup_mcp_servers, MCPServerInstance
 
+from dotenv import load_dotenv
+
+load_dotenv()
 
 SUPPORTED_TOOL_NAME: Final[str] = "web_search"
```
maximilien, line 33: I don't like how this is fixed. We need to remove and use the flexible tool support @akihikokuroda is adding

CC: @planetf1

src/maestro/cli/fastapi_serve.py
maximilien: maybe something like:

`enable_cors: true`

And false by default or w/o specifying

ajbozarth: I'm making an env var list that leave it disable when unset, currently testing it

ajbozarth:
Just pushed an update that limits the CORS to a allowed origin list configured by an env var (`CORS_ALLOW_ORIGINS`) and only the `GET` and `POST` method, rather than allowing everything.

I've tested this with my web app use case and confirmed that when `CORS_ALLOW_ORIGINS` is unset CORS behaves as default, that is, only allowing same origin. Then by setting `CORS_ALLOW_ORIGINS` in the  web apps's `.env` to the web app url it works as intended.

ajbozarth:
> Maestro Run Tests / test (pull_request)

TIL GitHub Actions does a merge with latest main before running CI rather than just running on the PR branch

PR #576, opened 2025-07-02
Workflow http

### LINKED ISSUE #569: feature: server workflow
This is adding same feature as #557 to workflow.

> Adds a maestro serve command to expose agents via an HTTP endpoint (557)

This is used by #556 / #513

akihikokuroda:
fix #569


docs/USER_GUIDE.md
```diff
@@ -195,6 +195,9 @@ The Maestro Command Line Interface (CLI) allows users to manage workflows that i
   - `--host HOST`: host to bind to (default: 127.0.0.1)
   - `--agent-name NAME`: specific agent name to serve (if multiple agents in file)
   - `--streaming`: enable streaming responses
+- `maestro serve-workflow` AGENTS_FILE WORKFLow_FILE [options]: serve workflow via HTTP API endpoints
```
psschwei, line 198: would it be possible to do like with `maestro run` and be able to pass either (1) just an agents file, or (2) both an agents file and workflow file? 

(I think it's a nicer UX to have just one command, even if we keep separate functions for both patterns behind the scenes)

akihikokuroda: I'll try it.  Thanks!

docs/USER_GUIDE.md
```diff
@@ -300,6 +301,57 @@ spec:
 - Consider adding authentication and rate limiting for production use
 - Use HTTPS in production environments
 
+### Serving Workflow via HTTP API
+
+The `maestro serve-workflow` command allows you to expose Maestro workflow as HTTP API endpoints, making them accessible via REST API calls. This is useful for integrating workflow into another workflows, web applications, microservices, or other systems that need to communicate with AI workflow.
+
```
maximilien, line 307: If we call this `serve-workflow` then for consistency we should call serve `serve-agent`

Maybe `serve` can support both agents and workflow with -an and -w respectively?

maximilien (approved):
LGTM 

See my one comment about CLI consistency

george-lhj:
pls give me 20 min to review this, will do it rn

src/maestro/cli/fastapi_serve.py
```diff
@@ -186,3 +186,108 @@ def serve_agent(
...
+        @self.app.post("/chat", response_model=WorkflowChatResponse)
+        async def chat(request: WorkflowChatRequest):
+            """Chat with the workflow."""
+            try:
+                if not self.workflow:
+                    raise HTTPException(status_code=500, detail="No workflow loaded")
+                
+                response = await self.workflow.run(request.prompt)
+                return WorkflowChatResponse(
+                    response=str(response),
```
ajbozarth, line 240:
```suggestion
                    response=json.dumps(response),
```
This need to be valid JSON for a front-end to parse it

ajbozarth:
I tested this with a front-end application and with the json dumps edit I suggested (as well as disabling CORs like I did in #574 ) it works great

docs/USER_GUIDE.md
ajbozarth: This is actually just a doc update issue, the command is already unified as `maestro serve` and determines if it's agent or workflow depending on the inputs (@akihikokuroda updated it per @psschwei review above)

george-lhj (approved):
see comment, otherwise lgtm

docs/USER_GUIDE.md
```diff
@@ -300,6 +301,57 @@ spec:
...
+```bash
+curl -X POST "http://127.0.0.1:8000/chat" \
+     -H "Content-Type: application/json" \
+     -d '{"prompt": "Hello, how are you?", "stream": false}'
+```
+
+Response:
+```json
+{
+  "response": "Hello! I'm doing well, thank you for asking. How can I help you today?",
```
ajbozarth, line 332: Also worth noting that if run with a workflow this is not what "response" looks like, it's a stringified json object

akihikokuroda:
@george-lhj   Thanks!

>As a result, for more complex workflows, perhpas we need to consider this and include it in the documentation to disable /enable certain tools

I'll address this in a separate PR (probably related to agent and/or tool related)

PR #578, opened 2025-07-03
enhancement: meta agent v2, agents file

george-lhj:
- updated meta agents using new prompts, models in order to produce more consistent results; specific to agents.yaml file curation
- removed extra print statement from step.py which I was fixing for debug earlier.

Example input: I want to fetch the current stock prices for Apple and Microsoft, and then analyze which one has performed better over the past week.

Example output:
```
🐝 Running TaskInterpreter...

🐝 Response from TaskInterpreter: <think>
Okay, so the user wants me to figure out how to fetch current stock prices for Apple and Microsoft and then analyze which one has performed better over the past week. Let's break this down.

First, I need to get the current stock data for both companies. That means using a financial data service or API because fetching stock prices isn't something I can do with basic functions. So, I'll probably need an agent that can pull real-time stock info.

Once I have today's data, I also need historical data going back at least a week. To compare performance over the past week, having daily averages would be useful. That way, I can calculate or compare trends easily.

After gathering both current and historical data, the next step is to analyze which stock has been performing better. This involves comparing their values or growth rates over those seven days. So, another agent that can perform this analysis makes sense.

Putting it all together, I think three agents are needed: one for current prices, one for historical averages, and a third for the performance comparison.
</think>

```
number of agents: 3
agent1: stock_data_retriever – Retrieves current stock prices for Apple (AAPL) and Microsoft (MSFT).
agent2: financial_historical_data_retriever – Retrieves historical stock data for the past week for both companies.
agent3: stock_performance_analyzer – Analyzes which stock has performed better over the past week based on historical data.
```

🐝 Running AgentYAMLBuilder...

🐝 Response from AgentYAMLBuilder: <think>
Okay, so I need to figure out how to create a YAML file with three agents based on the user's instructions. Let me break this down step by step.

First, the user provided an example where they had two agents and specified each agent's name, description, framework, mode, etc. Now, they're asking for three agents: stock_data_retriever, financial_historical_data_retriever, and stock_performance_analyzer.

I remember that in YAML, each agent is a separate block separated by `---`. Each agent has metadata with the name and labels, followed by the spec which includes framework, mode, description, and code for code agents or model, framework, mode, description, and instructions for LLMs.

Looking at the example, stock_data_retriever uses a code agent because it's doing data retrieval via API. So I'll need to write Python code that makes an HTTP request to get current prices for AAPL and MSFT. The code should import necessary modules like requests and json, fetch the data from the specified endpoints, parse the JSON response, extract the lastTradePrice and marketCap values, and assign them to variables.

Next, financial_historical_data_retriever is also a code agent but it needs to make API calls for historical data over the past week. I'll use pandas to create a date range starting from 7 days ago up to today. For each day in this range, I'll construct the URL with the appropriate endpoint and fetch the JSON data. Then, I'll extract the daily volume, close price, high, low, open, etc., into a DataFrame and save it as an Excel file.

The third agent is stock_performance_analyzer, which uses an LLM because it's analyzing data to determine performance. The description should explain its purpose briefly. The instructions need to be detailed enough for the model to understand how to compare the two stocks based on historical data, considering metrics like percentage change in price and volume analysis.

I also need to ensure that each agent is properly formatted with `|` separating multiline fields and that all code blocks are correctly indented without any function definitions or return statements. For the LLM agent, I have to make sure the instructions are clear on how to analyze the data from the Excel file.

Wait, in the example, the financial_historical_data_retriever uses pandas and requests but doesn't import anything else. So I should include necessary imports like pandas as pd and datetime for date handling. Also, when making API calls, sometimes you need to handle authentication or proxies, but since it's a mockup, maybe that's not required here.

Another thing is the time zone in the example output shows 2025-07-03T19:55:39+00:00. So when the user runs this code, they should be aware of the current time and date in ISO format.

I think I've covered all the points. Now, putting it all together into a single YAML file with each agent properly formatted.
</think>

```yaml
apiVersion: maestro/v1alpha1
kind: Agent
metadata:
  name: stock_data_retriever
  labels:
    app: generated
spec:
  framework: code
  mode: local
  description: |
    Retrieves current stock prices for Apple (AAPL) and Microsoft (MSFT).
  code: |
    import requests
    import json

    def get_current_stock_prices():
        url_aapl = "https://api Stocks.com/v1/marketstatus/AAPL"
        url_msft = "https://api Stocks.com/v1/marketstatus/MSFT"

        response_aapl = requests.get(url_aapl)
        response_msft = requests.get(url_msft)

        data_aapl = json.loads(response_aapl.text)
        data_msft = json.loads(response_msft.text)

        return {
            "AAPL": {"lastTradePrice": data_aapl.get("lastTradePrice", 0), 
                    "marketCap": data_aapl.get("marketCap", 0)},
            "MSFT": {"lastTradePrice": data_msft.get("lastTradePrice", 0),
                    "marketCap": data_msft.get("marketCap", 0)}
        }

    current_prices = get_current_stock_prices()
    print(f"Current stock prices as of {datetime.datetime.now().isoformat()}:")
    print(current_prices)

---
apiVersion: maestro/v1alpha1
kind: Agent
metadata:
  name: financial_historical_data_retriever
  labels:
    app: generated
spec:
  framework: code
  mode: local
  description: |
    Retrieves historical stock data for the past week for both companies.
  code: |
    import pandas as pd
    import datetime
    from requests import get

    def get_historical_data(ticker):
        start_date = (datetime.datetime.now() - datetime.timedelta(days=7)).isoformat()
        end_date = datetime.datetime.now().isoformat()

        url = f"https://api Stocks.com/v1/histdata?start={start_date}&end={end_date}&interval=daily&symbol={ticker}"

        response = get(url)
        data = json.loads(response.text)

        df = pd.DataFrame(data)
        df.to_excel("historical_data.xlsx", index=False)
        print("Historical data saved to historical_data.xlsx")

    get_historical_data("AAPL")
    get_historical_data("MSFT")

---
apiVersion: maestro/v1alpha1
kind: Agent
metadata:
  name: stock_performance_analyzer
  labels:
    app: generated
spec:
  framework: beeai
  mode: remote
  model: deepseek-r1:latest
  description: |
    Analyzes which stock has performed better over the past week based on historical data.
  instructions: |
    Analyze the historical data in historical_data.xlsx to determine which stock (AAPL or MSFT) has performed better over the past week. Consider metrics such as percentage change in price, volume trends, and market capitalization changes. Provide a clear comparison and recommendation based on these factors."
```

### DIFF
--- a/demos/workflows/meta-agents-v2/agents_file_generation/agents.yaml
+++ b/demos/workflows/meta-agents-v2/agents_file_generation/agents.yaml
@@ -1,0 +1,160 @@
+apiVersion: maestro/v1alpha1
+kind: Agent
+metadata:
+  name: TaskInterpreter
+  labels:
+    app: meta-agents-v2
+spec:
+  model: deepseek-r1:latest
+  framework: beeai
+  mode: remote
+  description: "Parses user goals and converts them into a structured list of agents with descriptions."
+  instructions: |
+    You are a **task-to-agent planner** that takes a user's natural language input and outputs a list of agents needed to accomplish that task.
+
+    ## Your task:
+    Convert the user’s goal into a **structured YAML block** that includes:
+    - The number of agents
+    - Agent names (short and descriptive)
+    - One-line description for each agent’s function
+
+    ## Output Format:
+    ```
+    number of agents: <X>
+    agent1: <name> – <description>
+    agent2: <name> – <description>
+    ...
+    ```
+
+    ## Guidelines:
+    - Use **concise, meaningful names** for each agent (e.g., `weather_retriever`, `trend_analyzer`).
+    - Descriptions should be **one sentence** starting with a verb, clearly stating the function.
+    - Prefer 1–3 agents unless task truly requires more.
+    - Try to combine related logic into a single agent when it simplifies the workflow without losing clarity.
+    - **Avoid duplicating tasks** across agents.
+    - If the user’s request is vague, make **reasonable assumptions** and err on the side of simplicity.
+
+    ## Examples:
+
+    ### Example 1:
+    Input: "I want to compare the current weather with the historical averages."
+    Output:
+    ```
+    number of agents: 2
+    agent1: weather_retriever – Retrieves today’s weather for a location.
+    agent2: trend_analyzer – Compares today’s weather to the 7-day historical average.
+    ```
+
+    ### Example 2:
+    Input: "Get the latest news headlines and summarize them."
+    Output:
+    ```
+    number of agents: 2
+    agent1: news_fetcher – Retrieves current news headlines from a news API.
+    agent2: summarizer – Condenses headlines into a short summary paragraph.
+    ```
+
+    ### Example 3:
+    Input: "Analyze sentiment of Reddit posts about Nvidia."
+    Output:
+    ```
+    number of agents: 2
+    agent1: reddit_scraper – Fetches Reddit posts about Nvidia.
+    agent2: sentiment_analyzer – Determines sentiment of the collected posts.
+    ```
+  tools:
+    - None
+
+---
+apiVersion: maestro/v1alpha1
+kind: Agent
+metadata:
+  name: AgentYAMLBuilder
+  labels:
+    app: meta-agents-v2
+spec:
+  model: deepseek-r1:latest
+  framework: beeai
+  mode: remote
+  description: "Generates valid agents.yaml files using either LLM or code agents depending on the task."
+instructions: |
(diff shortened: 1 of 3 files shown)


george-lhj:
as you can see, already we are getting pretty good results. now we need a way for us to take these reuslts, save its history/conversation, and make sure we can make some small tiny adjustments to it as needed

maximilien (approved):
Love the results I can see. Hopefully this meta agent could be used as a basis for the maestro builder backend agent to create the generic maestro yamls :)

Issue #273, opened 2025-02-27
Create tool YAML and list that can be populated from a tools catalog


maximilien:
As part of the work to test a meta-agent that can generate agents and workflows based on textual description. We hypothesize that access to a list of tools will be necessary since this could allow not only enhancing existing agents with tools for react-style agents but also ability for meta-agent to know what is missing or available to implement the workflow. This is related to issue #XXX.

Suggesting the following tasks:

- [ ] define tool YAML. Should follow existing YAML structure but also at minimum add: input and output schema for the tool and link or code for tool
- [ ] create JSON schema for tool
- [ ] include tool schema validation in current CI for schema validations
- [ ] test tool schema by manually creating a list from existing tools, e.g., BeeAI tools, CrewAI, and other tool repositories
- [ ] modify existing demos to ensure the react agents there use the tools in list
- [ ] test ability for meta-agent to use tool list

CC: @george-lhj @akihikokuroda

george-lhj:
Should also have ability to handle/create custom tools if possible, since for more complicated demos this is what primarily will be used

maximilien:
Add a task @george-lhj

maximilien:
Now that agents are connected to tools we can close this

PR #583, opened 2025-07-08
feature: add some beeai tool yamls

akihikokuroda:
This is a part of #273 

Adding some BeeAI tool yamls


maximilien (approved):
Love these @akihikokuroda 

Only question is why do they have to be "BeeAI" tools. Some could be generic for any framework, no?

akihikokuroda:
#273 has 
> test tool schema by manually creating a list from existing tools, e.g., **BeeAI tools**, CrewAI, and other tool repositories

So I just started with BeeAI tools.  I'll work with some MCP tools, next.

PR #591, opened 2025-07-09
enhancment: auto update releases names 

### LINKED ISSUE #589: feature: automate name release
Since we plan on creating more releases in the future, it might be helpful to automating the naming, based on the release names we have

george-lhj:
added functionality so each tag release a triggered, we read the next name from .release_names files, and then use that as our release names.

additionally, also auto creates a PR to strikethrough the name after release is published.

have tested its func on my fork

fix #589


george-lhj:
sorry @ajbozarth , realized that i accidentally deleted a few things. ready now

.github/workflows/maestro_release.yaml
```diff
@@ -71,7 +71,6 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           GITHUB_USER: ${{ github.actor }}
-
```
ajbozarth, line 74: nit: no need to delete

.github/workflows/maestro_release.yaml
```diff
@@ -80,9 +79,56 @@ jobs:
...
+      - name: Install dependencies
+        run: |
+          uv pip install -e .[crewai]
+
+      - name: Get release name
+        id: get_name
+        run: python tools/get_release_name.py
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GITHUB_REPOSITORY: ${{ github.repository }}
```
ajbozarth, line 104: IIUC we could move this to the build step and have that step output the value for later use below

.github/workflows/maestro_release.yaml
```diff
@@ -80,9 +79,56 @@ jobs:
...
+          python-version: "3.12"
+
+      - name: Install uv and activate the environment
+        uses: astral-sh/setup-uv@v6
+        with:
+          activate-environment: true
+
+      - name: Install dependencies
+        run: |
+          uv pip install -e .[crewai]
```
ajbozarth, line 97: We shouldn't need to checkout and build again, that's why we download the dist artifact, iiuc you only did this for the get_name step, which I believe we can move as I pointed out in the other comment

.github/workflows/maestro_release.yaml
```diff
@@ -80,9 +79,56 @@ jobs:
...
+      - name: Get release name
+        id: get_name
+        run: python tools/get_release_name.py
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GITHUB_REPOSITORY: ${{ github.repository }}
+
+      - name: Mock artifact for test release
+        run: |
+          mkdir -p dist && echo "test" > dist/fake.txt
```
ajbozarth, line 108: what is the purpose of this step?

.github/workflows/maestro_release.yaml
```diff
@@ -80,9 +79,56 @@ jobs:
...
+        uses: peter-evans/create-pull-request@v6
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          commit-message: "chore: strike out used release name '${{ steps.get_name.outputs.release_name }}'"
+          branch: strikeout-${{ github.ref_name }}-${{ github.run_id }}
+          base: main
+          title: "chore: strike out used release name '${{ steps.get_name.outputs.release_name }}'"
+          body: |
+            This PR was automatically created after tagging `${{ github.ref_name }}` to strike through the used release name.
+          labels: automated-pr
```
ajbozarth, line 134: I would argue that this should be in its own followup job. In addition if we're already opening a followup to release PR then we should also have it increment the version in pyproject.toml and README.md to the next expected version (I suggested this a followup in the previous PR)

akihikokuroda (approved):
LGTM.  Thanks!

ajbozarth (changes requested):
I listed a bunch of inline comments but I believe this could be better streamlined and separated.

.github/workflows/maestro_release.yaml
ajbozarth: fyi https://docs.github.com/en/actions/how-tos/writing-workflows/choosing-what-your-workflow-does/passing-information-between-jobs

.github/workflows/maestro_release.yaml
```diff
@@ -37,8 +39,16 @@ jobs:
           python-version: "3.12"
       - name: Install uv
         uses: astral-sh/setup-uv@v6
+      - name: Install dependencies
+        run: |
+          uv pip install -e .[crewai]
```
ajbozarth, line 44: this shouldn't be needed, the python and uv install steps should be enough to run the py script (if it's not then it should just move to the test step)

.github/workflows/maestro_release.yaml
george-lhj: moved the release name to the build and exposed it as output so that it can be used without having to recompute

.github/workflows/maestro_release.yaml
```diff
@@ -80,9 +88,33 @@ jobs:
         with:
           name: maestro-dist
           path: dist
+      - uses: actions/checkout@v4
```
ajbozarth, line 91: not needed

ajbozarth (changes requested):

.github/workflows/maestro_release.yaml
ajbozarth: if you open a followup then this part LGTM

.github/workflows/maestro_release.yaml
```diff
@@ -9,26 +9,28 @@ jobs:
...
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
+      - name: Install uv and activate the environment
+        uses: astral-sh/setup-uv@v6
+        with:
+          activate-environment: true
+      - name: Install dependencies
+        run: |
+          uv pip install -e .[crewai] requests
```
ajbozarth, line 23: why are we adding `requests` here?

.github/workflows/maestro_release.yaml
```diff
@@ -37,8 +39,16 @@ jobs:
           python-version: "3.12"
       - name: Install uv
         uses: astral-sh/setup-uv@v6
+      - name: Install dependencies
+        run: |
+          uv pip install -e . requests
```
ajbozarth, line 44: ah, you need `requests` for the script, in that case you don't have to install all of maestro just this will do
```suggestion
          uv pip install requests
```

.github/workflows/maestro_release.yaml
```diff
@@ -62,7 +64,7 @@ jobs:
     needs: build
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4 # TODO: replace with demo repo once it's created
+      - uses: actions/checkout@v4
```
ajbozarth, line 67: why delete the todo, it's still relevant

.github/workflows/maestro_release.yaml
```diff
@@ -46,10 +46,12 @@ jobs:
...
-        run: uv run python tools/get_release_name.py
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          GITHUB_REPOSITORY: ${{ github.repository }}
+        run: |
+          uv run env \
+            GITHUB_OUTPUT="$GITHUB_OUTPUT" \
+            GITHUB_TOKEN="${{ secrets.GITHUB_TOKEN }}" \
+            GITHUB_REPOSITORY="${{ github.repository }}" \
+            python tools/get_release_name.py
```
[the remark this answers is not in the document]
ajbozarth: ah, so it prevents some errors from being hidden when running the script, g2k

george-lhj:
> Love it… though how do we test it? I guess next release :)

in a fork, you can just tag any one of your branches and push the tag, it would trigger this release workflow

PR #594, opened 2025-07-09
chore: switch to `uv sync` for dev install

### LINKED ISSUE #592: switch to use `uv sync` for dev install
After working in another project I've found that the more accepted command for installing a uv project in development mode is `uv sync` instead of `uv pip install -e .`

We should look into updating the docs, scripts, and CI to use `uv sync` wherever we currently use `uv pip install -e .`.

The only limitation here would be that `uv sync` is actually equivalent to `uv pip install -e .[crewai]` as it always installs all optional dependencies, so we would need to check if there are place where `uv pip install -e .` is still appropriate.

ajbozarth:
Fixes #592 (See there for the reasoning)

Replaces `uv pip install -e .` with `uv sync` in the docs, scripts and CI.

Also fixes a bug in `pyproject.toml` that only surfaces when using `uv sync` and not `uv pip install -e .` or `uv build`

### DIFF
--- a/.github/workflows/maestro_check-mermaid.yaml
+++ b/.github/workflows/maestro_check-mermaid.yaml
@@ -21,7 +21,7 @@ jobs:
         activate-environment: true
     - name: Install dependencies
       run: |
-        uv pip install -e .
+        uv sync
     - name: Check mermaid
       run: |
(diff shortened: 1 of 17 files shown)


maximilien (approved):
Love this. Simplify simplify simply

Thanks @ajbozarth

PR #602, opened 2025-07-10
demo: make qchem ai public and add readme

AngeloDanducci:


maximilien (changes requested):
Can you integrate into CI or male sure it's executing?

AngeloDanducci:
> How is your summary performing? I remember one issue I had is that with very large papers the summaries weren't super good, meaning they can't capture the entire paper because of its token limit. And also, I think I switched away from llama3.1

Varies quite a bit run to run on my local machine - the entire papers are definitely not captured in the summary.

What model did you switch to, granite?

AngeloDanducci:
@maximilien did a fresh maestro install and then followed the README - working on local machine.  Do we have CI set up for demos?  I see a simple demo in CI but nothing in the tests dir or CI in general for all demos.

maximilien (approved):
LGTM

PR #606, opened 2025-07-11
enhancement: connect Opik dashboard

george-lhj:
makes it easier to log traces, see scores, and understand reasoning behind a score


george-lhj:
> CN we get LLM as a judge for prompt?
> 
> Otherwise good start

Actually, Opik already uses LLM as a judge to create our relevancy and hallucination scores. So this is already impelemtented, and why I defined the ScoringAgent with a model which requires Ollama backend

Issue #603, opened 2025-07-10
feature: support `maestro serve` in containerized maestro


ajbozarth:
A follow up to #546 which added the ability to serve maestro agents and workflows as a HTTP API.

We want to do the same with a containerized instance of maestro. That is we should be able to deploy a container that provides the `maestro serve` API

ajbozarth:
I was thinking specifically of the `maestro serve` command itself. So the `docker run` would take agent and workflow yawls as input, or have them as them as input when building the image and then serve the yawls that the image contains. The former would be more of a publishable image for use and the later would be a usable dockerfile for devs to leverage to create custom images. I'm unsure which would be the better solution (or perhaps both)

akihikokuroda:
oh, I see.  I misunderstood.  The former doesn't require the use installing maestro at all.  It's good for trial.   I think there are 3 steps to solve this issue.

1. make a base container image contains maestro
2. create another image for the former case by adding the entrypoint for cmd to the base container image
3. document how to create the custom maestro image using the base container image

ajbozarth:
Though #610 has been merge there is still open tasks:

- add documentation on extending the image
- update the releaser to build images for both `linux/amd64` and `linux/arm64` platforms and created a manifest for accessing them (I did this in a previous project: https://github.com/open-quantum-safe/oqs-demos/blob/main/.github/workflows/push-manifest.yml)

In addition the current Dockerfile is broken for me.

If we want I can open up follow up issues for the above and we can close this. I already have some of the fixes done locally, but I'm still hitting errors

ajbozarth:
As the PR #626 and followup issue #625 cover the remaining tasks I'm closing this

PR #610, opened 2025-07-11
enhancement: add maestro Dockerfiles

akihikokuroda:
part of https://github.com/AI4quantum/maestro/issues/603

This PR has 3 Dockerfile

maestro base image: it has the runnable maestro image.
maestro cli image: it add maestro as the entry point
maestro custom image: it is an example for customizing the maestro image.


Dockerfile-base
```diff
@@ -0,0 +1,23 @@
...
+ENV PYTHONUNBUFFERED=1
+
+# Install pip and dependencies
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends \
+    git \
+    && rm -rf /var/lib/apt/lists/*
+
+# Install dependencies
+RUN pip install git+https://github.com/AI4quantum/maestro.git@v0.2.0
```
ajbozarth, line 15: is there a way to make the version tag customizable at build so we don't need to update this file every release? like it can default to `main` unless a tag is specified?

Dockerfile-base
```diff
@@ -0,0 +1,23 @@
...
+
+RUN chown -R 1000:100 /usr/src/app &&\
+    mkdir -p /usr/src/app/src/media && chown 1000:100 /usr/src/app/src/media
+
+EXPOSE 8000
+USER 1000:100
+
+# ENTRYPOINT ["maestro"]
```
ajbozarth, line 23: is there a reason to comment this out? could we uncomment this and not need the separate `Dockerfile-cli` below?

Dockerfile-custom
```diff
@@ -0,0 +1,9 @@
+FROM localhost/maestro-base:latest
+
+# COPY <src> agent.yaml      # copy agent file into image
+# COPY <dest> workflow.yaml  # copy workflow file into image
+
+CMD ["serve", "agent.yaml", "workflow.yaml", "--port=5000"]  # for workflow serve
+#CMD ["serve", "agent.yaml", "--port=5000"]                  # for agent serve
```
ajbozarth, line 7: why change the port isn't the default 8000 correct based on the base image? 

also I found in my previous work that when running from docker we need to set `"--host", "0.0.0.0"` as well to handle outside connections

Dockerfile-custom
```diff
@@ -0,0 +1,9 @@
...
+
+ENTRYPOINT ["maestro"]
```
ajbozarth, line 9: IIUC this would run `ENTRYPOINT CMD` when docker run?

So this could work out of the box if placed in a dir with a agent.yaml and workflow.yaml

It would also allow direct use by passing in different args to override CDM correct?

ajbozarth: Also if we uncomment the ENTRYPOINT in the absence of image would we even need this line?

ajbozarth (commented):
Thanks for working on this, it was exactly what I was thinking of.

I don't fully understand the reason for all the parts of Dockerfile-base like the chown and USER, but I have a handful of clarifying comments throughout.

ajbozarth:
Also an idea on naming: if we're able to combine `Dockerfile-base` and `Dockerfile-cli` like I implied in my comments we could just rename that to `Dockerfile` in which case I think `Dockerfile-custom` might be better named `Dockerfile.example` or even just moved into a code block in a doc

Dockerfile-base
ajbozarth: or would it make sense to do a dev install here from the current directory since in theory this would be built from the root of the repo

ajbozarth: Also once this is merged it would be worth creating a followup task to publish it to ghcr as part of the release CI

Dockerfile-custom
```diff
@@ -0,0 +1,9 @@
+FROM localhost/maestro-base:latest
+
+# COPY <src> agent.yaml      # copy agent file into image
+# COPY <dest> workflow.yaml  # copy workflow file into image
```
psschwei, line 4: I would mount these as a volume at runtime
as an example of customization, maybe show to add other python libraries (e.g. pip install ...)

akihikokuroda:
@ajbozarth The base images can be the base of other images in addition to the cli use.  It can be used in some other cases, such as `maestro deploy k8s`, operator instrall, etc..  It make sense to have the independent image as the base.  WDTY?

I'm still working for the custom/example Dockerfile and the image upload CI.

maximilien (commented):
Why is this a DRAFT? What's missing @akihikokuroda ?

akihikokuroda:
I'm adding upload CI and Docs.

.github/workflows/maestro_build_upload_images.yaml
```diff
@@ -0,0 +1,40 @@
...
+        with:
+          name: maestro-dist
+          path: dist/*
+      - name: Build and push the image
+        run : |
+          VERSION=$(grep -E '^(version|tool\.poetry\.version) *= *"[^"]+"' "$PYPROJECT_TOML" | \
+            head -n 1 | \
+            sed -E 's/.*"([^"]+)".*/\1/')
+          uv run tools/buidimg.sh	  
+          docker login --username akihikokuroda --password "${{ secrets.GITHUB_TOKEN }}" ghcr.io
```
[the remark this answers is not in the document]
ajbozarth: here's an example I wrote for another project https://github.com/open-quantum-safe/oqs-demos/blob/main/.github/workflows/push-manifest.yml

maximilien (approved):
LGTM

PR #617, opened 2025-07-14
bug: fixes issue #616

maximilien:

Issue #585, opened 2025-07-08
bug: crewAI tests fails and instruction does not work


maximilien:
```bash
___________________________________________ CrewAITest.test_agent_runs ___________________________________________

self = <crewai_agent.test_crewai.CrewAITest testMethod=test_agent_runs>

    def test_agent_runs(self) -> None:
        agents_yaml = parse_yaml(os.path.join(os.path.dirname(__file__), "agents.yaml"))
        workflow_yaml = parse_yaml(
            os.path.join(os.path.dirname(__file__), "workflow.yaml")
        )

        try:
            workflow = Workflow(agents_yaml, workflow_yaml[0])
        except Exception as excep:
            raise RuntimeError("Unable to create agents") from excep

>       result = asyncio.run(workflow.run())
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^

tests/agents/crewai_agent/test_crewai.py:31:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
../../.pyenv/versions/3.12.8/lib/python3.12/asyncio/runners.py:194: in run
    return runner.run(main)
           ^^^^^^^^^^^^^^^^
../../.pyenv/versions/3.12.8/lib/python3.12/asyncio/runners.py:118: in run
    return self._loop.run_until_complete(task)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
../../.pyenv/versions/3.12.8/lib/python3.12/asyncio/base_events.py:686: in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
src/maestro/workflow.py:59: in run
    self._create_or_restore_agents()
src/maestro/workflow.py:95: in _create_or_restore_agents
    agent_instance = cls(agent_def)
                     ^^^^^^^^^^^^^^
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <maestro.agents.crewai_agent.CrewAIAgent object at 0x116c45310>
[...]

PR #620, opened 2025-07-14
refactor: remove maestro bash script

### LINKED ISSUE #601: enhancement: remove `maestro` bash script
With the switch to uv and pip installable distributions we should delete the old `maestro` bash script at the root of the project and make sure all references to it in scripts use either `uv run maestro` if running a dev install or just `maestro` if pip installed.

The script can cause errors or confusion in some scenarios:

If a user has cloned the repo, but then pip installs maestro (not a dev install) the script will interfere when calling maestro at the root dir:
- If you *have not* done a previous dev install you'll get an error: `No module named maestro.cli.maestro`
- If you *have* done a dev install previously then it will use the dev install instead of the pip install as expected.

ajbozarth:
Fixes #601


ajbozarth:
Passes pytest locally, pushed early to check if the CI all passes as well. If CI passes this *should* correctly remove references to the bash script

ajbozarth:
> CI is failing. Please fix and I'll review when green. Thx

I had to include the changes from #617 to get the CI scripts to pass, but it should be green this time.

maximilien (commented):
LGTM but since you changed some files that @akihikokuroda created might be worth to get his OK first

maximilien:
My problem is that it did not work for me.

ajbozarth:
> My problem is that it did not work for me. The suggestion listed does not work. At least in my environment

I'm a bit confused, was this comment meant for our conversation in #585 ? The only feedback you have given on this PR was for @akihikokuroda to do a review and to wait for CI to pass.

akihikokuroda:
I have tried.  It seems working for mw.

PR #621, opened 2025-07-14
enhancement: connect frontend w API and meta agents

george-lhj:
- Connected API with meta agents.
- Allows modification w/ chat history (ex, once file generated, can ask to rename an agent, and it will do it)

Followup enhancements:

1) need to add modify stop.sh to kill ports that are ran (like 5174, which is frontend, 8001, which is the frontend api)
2) need to modify UI output, so that its easier to read the output. Everything is squished (the output is correct, just can't read it properly/easily)
3) once agent file generated, the right side yaml panel disappears
4) design decision

Currently these folders
agent_file_generation: builds agents.yaml
workflow_file_generation: builds workflow.yaml

Each workflow required a separate maestro serve instance and could not be hosted concurrently due to the single-port limitation (default: 8000), which means a user cant run both at the same time.

Need to either allow support for multiple, or combine current agent and workflow_file_generation files into one (which is possible)

### DIFF
--- a/.github/workflows/maestro_release.yaml
+++ b/.github/workflows/maestro_release.yaml
@@ -64,7 +64,7 @@ jobs:
     needs: build
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4 # TODO: replace with demo repo once it's created
+      - uses: actions/checkout@v4
         with:
           repository: AI4quantum/maestro-demos
       - uses: actions/download-artifact@v4
--- a/.github/workflows/maestro_operator-test.yaml
+++ b/.github/workflows/maestro_operator-test.yaml
@@ -52,8 +52,10 @@ jobs:
     - name: Deploy test agents, workflow, configmap and workflowrun
       run: |
         export PYTHONPATH=$PYTHONPATH:$(pwd)/src
-        maestro create-cr https://raw.githubusercontent.com/AI4quantum/maestro-demos/refs/heads/main/workflows/weather-checker.ai/agents.yaml
-        maestro create-cr https://raw.githubusercontent.com/AI4quantum/maestro-demos/refs/heads/main/workflows/weather-checker.ai/workflow.yaml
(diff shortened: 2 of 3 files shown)


maximilien:
Ok let me try and take a look tomorrow. For UI changes let's plan that after we move builder to its own repo.

george-lhj:
Screenshot 2025-07-15 at 3 44 07 PM
0" />

outputs correct yaml, UI is just formatted wrong

maximilien:
> Screenshot 2025-07-15 at 3 44 07 PM Screenshot 2025-07-15 at 3 44 35 PM
> outputs correct yaml, UI is just formatted wrong

Can we fix UI for this before moving @george-lhj ?

.github/workflows/maestro_build_upload_images.yaml
maximilien, line 1: Why are we adding this change as part of this PR? I would suggest separate.

.github/workflows/maestro_release.yaml
maximilien, line 1: Same here. This should be different PR

CONTRIBUTING.md
maximilien, line 1: Again separate PR so things are focused.

maximilien (changes requested):
@george-lhj see my initial comments. Let's separate the Docker changes with the builder. Reasoning is as we move builder code to its own repo these will need to change and move so why not do any build release stuff there. 

On this repo I just want to have a working vesion of builder so we can tag and move.

WDYT?

CONTRIBUTING.md
george-lhj: i didn't add this change. its pulling the latest changes from main, because I need your ./stop update, and when i pull it will take in things like docker changes

maximilien:
OK so now rebase and should be good.

george-lhj:
ok i've rebased, and then also added a separate PR to fix all parsing yaml errors in #635. 


This PR should be merged first

PR #635, opened 2025-07-16
enhancment: Update UI

george-lhj:
updated UI to parse and display generated yamls correctly.
Screenshot 2025-07-16 at 3 22 48 PM


george-lhj:
#637

PR #623, opened 2025-07-15
bug: Update start stop to Support Dedicated Builder API and Frontend

george-lhj:
Isolated Builder Backend from maestro serve: Assigned port 8001 to run api/main.py (Builder backend), allowing maestro serve to continue using port 8000.
- ./start.sh needs to point to overall maestro venv to support backend reqs also

Updated stop.sh:
- Gracefully shuts down Builder backend (8001) and frontend (5174).

PR #636, opened 2025-07-16
enhancement: make request data and response object configurable in RemoteAgent

### LINKED ISSUE #633: enhancement: make request data and response object configurable in RemoteAgent
currently remote agent allows an agent to call a given API URL with a prompt. This is potentially useful for 3rd party agent workflows like langflow, but the current implementation hard-codes the request data object and response object.

The request data is `{"prompt": prompt}` and the response expects a json object with a `response` key. This is not very useful since not all agent api's don't use that data and response format.

We should add the ability to configure these values in the agent yaml in some way

ajbozarth:
Fixes #633 

currently remote agent allows an agent to call a given API URL with a prompt. This is potentially useful for 3rd party agent workflows like langflow, but the current implementation hard-codes the request data object and response object.

The request data is `{"prompt": prompt}` and the response expects a json object with a `response` key. This is not very useful since not all agent api's don't use that data and response format.

This updates the spec to accept `request_template` and `response_template` to determine the json data object sent in the request and where to find the answer in the response object respectively.

An example agent yaml for a langflow api would look like:

```yaml
apiVersion: maestro/v1alpha1
kind: Agent
metadata:
  name: langflow
spec:
  framework: remote
  url: "http://127.0.0.1:7860/api/v1/run/721d7ad0-890b-4c9a-8541-92c80ab21de7"
  request_template: >
    {
      "input_value": "$prompt",
      "output_type": "chat",
      "input_type": "chat",
      "session_id": "user_1"
    }
  response_template: $response["outputs"][0]["outputs"][0]["results"]["message"]["text"]
```

### DIFF
--- a/src/maestro/agents/remote_agent.py
+++ b/src/maestro/agents/remote_agent.py
@@ -1,5 +1,7 @@
 #! /usr/bin/env python3
 # SPDX-License-Identifier: Apache-2.0
+import json
+from string import Template
 
 import dotenv
 import requests
@@ -24,6 +26,8 @@ class RemoteAgent(Agent):
         """
         super().__init__(agent)
         self.url = agent["spec"]["url"]
+        self.request_template = agent["spec"]["request_template"]
+        self.response_template = agent["spec"]["response_template"]
 
     async def run(self, prompt: str) -> str:
         """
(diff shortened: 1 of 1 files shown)


maximilien (changes requested):
Nice but please get 👍 from @akihikokuroda and / or Paul

maximilien:
Once this is merged and #623 then I will move builder and cut a release. Hopefully tomorrow early AM pacific.

Cc: @akihikokuroda @psschwei @george-lhj

maximilien (approved):
LTGM

Issue #639, opened 2025-07-17
chore: move builder to new repo


maximilien:

maximilien:
this was completed with PR #640

PR #640, opened 2025-07-17
chore: move builder

maximilien:
When merged this will fix issue #639


maximilien:
OK moved builder. As expected should not have caused any issues though new repo https://github.com/AI4quantum/maestro-builder will need some work.

So @george-lhj and @ajbozarth give me a quick review so I can merge and then cut a release here. 

Then @george-lhj and I will work on https://github.com/AI4quantum/maestro-builder and get it working with maestro as an installed package.

ajbozarth (approved):
LGTM, this was a much cleaner removal than demos was

PR #642, opened 2025-07-17
bug: fix release CI and prep patch release

ajbozarth:
The release CI had a couple bugs that were hit when `v0.3.0` was cut (https://github.com/AI4quantum/maestro/actions):

- the docker image release assumed the user/org name was all lowercase, AI4quantum is not, so a "to lower" step needed to be added (the step used was the recommended one by the community)
- the GitHub releaser action we used needs write permissions to cut a release, by default a user has write permissions on their own repos (like forks). When releasing on an org though the perms need to be explicitly given

I also incremented the version to 0.3.1 so we can cut a `v0.3.1` tag once this is merged and the release process should work as intended

### DIFF
--- a/.github/workflows/maestro_build_upload_images.yaml
+++ b/.github/workflows/maestro_build_upload_images.yaml
@@ -35,19 +35,22 @@ jobs:
           activate-environment: true
       - name: Build Maestro
         run: uv build
+      - name: get docker-compatible GITHUB_ORG
+        run: |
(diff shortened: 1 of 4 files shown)


ajbozarth:
sanity check release on my fork worked as intended when cutting a `v0.3.1` release on this branch. Once we merge this we just need to create a `v0.3.1` tag and push it and everything should be green

Issue #650, opened 2025-07-19
bug: beeai-agent and crewai dependency conflict


akihikokuroda:
Package name change duckduckgo-search to ddgs caused an issue in beeai-framework.   The changes is in python_v0.1.31 release.  The release has conflicting dependency with the crewai.

akihikokuroda:
Removing crewai that conflicts with beeai-framework

planetf1:
Agree - the crewai support was implemented early-on and effectively called existing crewai agent code. It did not create an agent based on crewai as is the case for beeai and openai. It should be removed and if/when we need crewai support we can adopt a similar pattern to that done for openi, dspy etc.

akihikokuroda:
At this moment, beeai-framework and crewai don't have conflict.  We can put crewai back in.

PR #651, opened 2025-07-21
bug: remove crewai agent 

akihikokuroda:
Remove crewai optional dependency from pyproject.toml.  It is causing a conflict with beeai-framework.

#650 left open to fix or clean up the CrewAI agent.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -22,7 +22,7 @@ dependencies = [
     "openapi>=2.0.0",
     "openai-agents[litellm]>=0.0.14",
     "pycron>=3.1.2",
-    "beeai-framework>=0.1.17",
+    "beeai-framework>=0.1.31",
     "slack_sdk>=3.35.0",
     "nest-asyncio>=1.6.0",
     "pydantic-ai[logfire]>=0.1.8",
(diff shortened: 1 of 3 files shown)


ajbozarth (changes requested):
If we do this, we would need to remove all the crewai uses and references from our code as well.

Have we looked into whether we can just unpin litellm? From a quick look, the lock files on main for both crewai and beeai-framework are using litellm `1.74.4`

Edit: if we do unpin it we should just remove it from pyproject.toml entirely since we don't use it directly in maestro

pyproject.toml
```diff
@@ -31,12 +31,7 @@ dependencies = [
     "uvicorn[standard]>=0.24.0",
     "dspy>=2.6.27",
     "kubernetes>=33.1.0",
-]
-
-[project.optional-dependencies]
-crewai = [
-    "crewai",
-    "litellm==1.67.5"
+    "ddgs>=9.4.0",
```
ajbozarth, line 34: separate from my other comment, shouldn't this replace `duckduckgo-search` rather than be added as a new line?

ajbozarth:
> Have we looked into whether we can just unpin litellm? From a quick look, the lock files on main for both crewai and beeai-framework are using litellm 1.74.4

Just did a check on this and it seems that though the litellm issue was solved now there a worse conflict with `json-repair`, so we should just continue with the removal of crewai

ajbozarth (approved):
LGTM as long as we open a followup to properly remove all crewai content

PR #653, opened 2025-07-21
chore: update locations for repos

george-lhj:
Our main readme shouldn't have instructions about a different repo, pointed the correct links for user to navigate to in case they want to learn more


maximilien (approved):
LGTM

PR #658, opened 2025-07-23
chore: add latest tag for docker images

### LINKED ISSUE #654: chore: include a `latest` tag
When I go to do things like install the demos, I shouldn't have to worry about which version I install. So being able to do something like `pip install git+https://github.com/AI4quantum/maestro.git@latest` would be nice (and similar to how PyPi works, as `pip install maestro` would return latest version)

ajbozarth:
Fixes #654

pushes a `latest` tag as well a a `#.#.#` tag for the docker images when doing a release.


psschwei (approved):
I think it's fine to hold off until next release

Issue #659, opened 2025-07-23
feature: add mcp tool support to agents in kubernetes cluster


akihikokuroda:
MCP server for tools defined in MCPTool yaml file are deployed in the kubernetes cluster now.

Maestro agent should use these MCP servers.

akihikokuroda:
Closed via #660

PR #660, opened 2025-07-23
chore: Agent MCP tool support utilities 

akihikokuroda:
Add MCP tool support utilities and DSPy agent change as the first example of the utility usage. 

ref #659

### DIFF
--- a/.github/workflows/maestro_mcp_tool_test.yaml
+++ b/.github/workflows/maestro_mcp_tool_test.yaml
@@ -83,9 +83,9 @@ jobs:
     - name: Run a workflow
       run: |
         # Install and start mockllm
-        pip install mockllm
+        uv pip install mockllm
         cp tests/utils/responses.yml responses.yml	
-        mockllm start --responses responses.yml
+        uv run mockllm start --responses responses.yml
 
         # Run workflow
         uv run maestro run tests/yamls/agents/openai_mcp_agent.yaml tests/yamls/workflows/openai_mcp_workflow.yaml
--- a/.github/workflows/maestro_mcp_tool_test.yaml
+++ b/.github/workflows/maestro_mcp_tool_test.yaml
@@ -88,6 +88,7 @@ jobs:
         uv pip install .
         cp ../tests/utils/responses.yml responses.yml	
         mockllm start --responses responses.yml &
+	sleep(10)
 
         # Run workflow
         uv run maestro run tests/yamls/agents/openai_mcp_agent.yaml tests/yamls/workflows/openai_mcp_workflow.yaml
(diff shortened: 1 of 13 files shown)


maximilien (changes requested):
LGTM except for lack of tests

CC: @ajbozarth since he should review this too

akihikokuroda:
So far the mcp tool access is enabled in BeeAI agent, OpenAI agent and DSPy agent. 
I'll add test for them next.

src/maestro/agents/beeai_agent.py
```diff
@@ -288,6 +290,23 @@ async def _create_agent(self):
             tools.append(sandbox_tool)
             self.print(sandbox_tool.name)
 
+        for tool in self.agent_tools:
+            if tool.lower() not in [
```
ajbozarth, line 294: rather than creating a giant list here we should save the above lists as vars and use a concat of the lists here instead, that would only define each tool name once.

Also it may be worth revisiting those hard-coded tool names. I originally added those as hard-coded to match the various names I found in both our test/demos/doc and the ones in beeai, but it may be better to have one documented name per internal tool to minimize potential overlap with map tool names

src/maestro/agents/openai_mcp.py
```diff
@@ -110,3 +111,35 @@ async def setup_mcp_servers(
...
+                if os.getenv("KUBERNETES_POD") == "true":
+                    url = service_url
+                if os.path.exists(
+                    "/var/run/secrets/kubernetes.io/serviceaccount/token"
+                ):
+                    url = service_url
+
+                # print("###")
+                print(tool_name)
+                print(url + "/sse")
```
ajbozarth, line 136: should these prints be cleaned up? they look like temp dev prints

ajbozarth (commented):
A couple minor comments on just a code read through. I haven't tried it locally, do we have some example demos that use the new tool yamls yet that I could try?

maximilien (changes requested):
Can we put some examples in issue comments and README or docs?

akihikokuroda:
I found major issues in this PR.  I'll do rework.

.github/workflows/maestro_mcp_tool_test.yaml
```diff
@@ -0,0 +1,111 @@
...
+    - uses: actions/checkout@v4
+    - name: Install Kind
+      run: |
+        curl -Lo ./kind "https://github.com/kubernetes-sigs/kind/releases/download/v0.20.0/kind-$(uname)-amd64"
+        chmod +x ./kind
+        sudo mv kind /usr/local/bin
+        which kind
+    - name: Create Kind cluster
+      run: |
+        kind create cluster --config tests/integration/deploys/kind-config.yaml
```
ajbozarth, line 24: Did you check if this could be done with a published action? https://github.com/marketplace/actions/kind-cluster

ajbozarth (commented):
one question on the CI, but my previous review has been addressed. I have not tested this locally though

.github/workflows/maestro_mcp_tool_test.yaml
akihikokuroda: Yes, I did.  it doesn't have the necessary config option.

PR #664, opened 2025-07-24
feat: enable llm url override via yaml spec

### LINKED ISSUE #657: feat: create a unified config for model url across agents
Currently the model url (often ollama) is configured differently for each agent. Eg. openai uses an env var, dspy uses a yaml config, beeai uses an env var with a default value

We should configure this with a yaml config to be used by all relevant agents, with an optional fallback for the old method

ajbozarth:
Fixes #657 

This removes the reliance on env vars for beeai and openai agents by allowing the config of the api url via the yaml and not just the env var and setting a dummy key by default if the api key env var is not set.

### DIFF
--- a/tests/yamls/agents/simple_local_agent.yaml
+++ b/tests/yamls/agents/simple_local_agent.yaml
@@ -6,7 +6,7 @@ metadata:
     app: test-example
 spec:
 
-  model: llama3.1
+  model: ollama/llama3.1
   framework: beeai
   mode: local
   description: this is a test
@@ -24,7 +24,7 @@ metadata:
   labels:
     app: test-example
 spec:
-  model: llama3.1
+  model: ollama/llama3.1
   mode: local
   framework: beeai
   description: this is a test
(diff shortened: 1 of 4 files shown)


ajbozarth:
once this is merged we will need to update the demos that use beeai local mode to include the provider in their model name,

in addition I still need to thoroughly test this, I plan to make the necessary updates in demos as a test and will open a PR there once I've finished my testing

psschwei:
> Kinda, the previous implementation was hard-coded to use local Ollama, so the only way to stay back-compatible would be to add a manual check and hard-code to ollama if the provider is missing. I could add that if we want to not break compatibility

I guess it would depend on if it's less work to do that than update the demos... :smile:

tests/yamls/agents/simple_local_agent.yaml
```diff
@@ -6,7 +6,7 @@ metadata:
     app: test-example
 spec:
 
-  model: llama3.1
+  model: ollama/llama3.1
```
psschwei, line 9: should one of the tests remain without the ollama (so that it tests what happens when no prefix)?

psschwei (approved):

PR #669, opened 2025-07-28
bug: fix beeai local issue

akihikokuroda:
Fix issue accessing to the url value in the agent yaml file.


akihikokuroda:
`self.agent["spec"]` doesn't exit.  It caused exception.

PR #670, opened 2025-07-28
fix: beeai agent broken from #664

ajbozarth:
This is a followup to #664 and #669 

#664 was merged without through testing and #669 partially fixed missed issues

This addresses the remaining bugs discovered when testing the edge cases introduced in #664 

- remote had bad urls (`self.base_url` already includes `/v1`)
- `from_name` spilts on `:` not `/` and thus requires a bit more work to still support the OllamaChatModel fallback


george-lhj (approved):
this works (I think), check slack comments

akihikokuroda (approved):
LGTM

PR #671, opened 2025-07-28
feat: support url or file path for code or instructions

### LINKED ISSUE #668: feature: allow code agent to use a python file instead of inline code
You should be able to pass in a local file path or url to a python file in addition to including the python code inline in the yaml. This would allow support for more verbose python agents and better code checking

ajbozarth:
Fixes #668 

This allows the values of `spec.code` and `spec.instructions` to point to a url or file path that holds the content in addition to including the value inline.

This does minimal validation:
- If the content is a list it assumes it's code block (currently only used in tests)
- It will try to fetch from a url if the value is in a url format, whether or not the url is valid
    - if the url is a GitHub or gist url it will convert it to a raw url if it is not already one
- It will only try to open a file if it exists, even if the file path is valid, meaning a malformed file path would be considered code
    - if the path is relative it will be treated relative to the agent.yaml file (not the maestro exec directory)
    - in order to support this the [`parse_yaml`](https://github.com/AI4quantum/maestro/blob/492374d839db7eb1f594e3e2a1a46da2bf466547/src/maestro/cli/common.py#L25) util now adds a top level value `source_file` containing the inputted yaml file path.


maximilien (changes requested):
Put some examples in issue and docs. That should be first thing to do in order to discuss and make sure you don't spend time implementing and finding out there are better / alternative approaches. Thx

ajbozarth:
> Put some examples in issue and docs 

I'll do that, this was tested using our internal example so I can add some docs/examples when adding the unit tests

ajbozarth:
> could we add tests for the new utility functions?

@psschwei tests added

> Put some examples in issue and docs

@maximilien I have an example in [draft pr on the internal use-case](https://github.ibm.com/Quantum-plus-AI/q-ai-platform/pull/41) but am still brainstorming how to include it in the docs or tests here. Once this is merged we should update some of the demos to use it though. If you want I can make the demo updates and open a draft PR there as an example

maximilien:
My point @ajbozarth is we should dicuss these things here first before code and before PR merged. Since you have code. Let's discuss. If all code and no code changes then let's merge.

We cannot shortcut this ... Thanks.

psschwei:
bringing this over from another discussion:

> file paths are relative to where maestro is run

that would mean that I would have to be in the directory with the source files to deploy the agent/workflow, right? 

for example, with this structure:

```bash
my_awesome_workflow
├── scripts
│   └── agent_code.py
└── workflows
    └── chat
        ├── agents.yaml
        └── workflow.yaml
```

and a code agent that was using `scripts/agent_code.py` as the filename this would only work if I was in the `my_awesome_workflow` directory? And not from my home directory and running `maestro serve my_awesome_workflow/workflows/chat/agents.yaml my_awesome_workflow/chat/workflow.yaml` ?

I'm not sure I like that.

psschwei:
> common issue with using relative file paths

would it be worth having an extra flag in the maestro command to pass the directory where the source files are? 

open to other ideas too, but having the maestro command be dependent on the pwd feels like a bad UX to me (see also #506)

ajbozarth:
> open to other ideas too, but having the maestro command be dependent on the pwd feels like a bad UX to me

I did a little digging on what is considered the "acceptable" way to handle paths in app configs and in general the path is relative to the config file (in our case agent.yaml) or enforces use of absolute paths. If this is how we want to handle this it would require a bit of work since we read in the yaml files using a utility function in many places in the code and would need to update all of them to save the location of the agent yaml

> would it be worth having an extra flag in the maestro command to pass the directory where the source files are?

I'm not sure I like this idea, but it would require only slightly less code changes than my above idea.

psschwei:
let's let @maximilien @akihikokuroda @george-lhj weigh in too

ajbozarth:
> I think i would prefer absolute paths

What do you mean by this? that we should enforce absolute paths? or just that by practice we recommend absolute paths? This implementation supports both, but absolute paths are difficult to use when sharing yamls via a git repo. 

Also another reminder that this feature does not replace the current functionality, so there is no reason to change our demos. This just give more flexibility

ajbozarth:
> My main concern was just around making the behavior explicit in the docs so users know how paths are resolved (e.g., relative to current working directory, not the YAML file location)

I'll add this to the docs alongside an example yaml like the one I included above.

>  If thats already addressed then it shouldn't affect the other repos and should be fine

Does that mean that as long as I document it you're ok with the current implementation?

ajbozarth:
> Yes sure, that would be good

I assumed you meant that I should add docs about code agents, so I've done so

src/maestro/agents/utils.py
```diff
@@ -0,0 +1,36 @@
...
+def is_file(text):
+    try:
+        return Path(text).exists()
+    except OSError:
+        return False
+
+
+def get_content(text):
+    if text is None:
+        return None
```
akihikokuroda, line 28: Do we have list in `code` or `instructions`?

ajbozarth: a multiline value is read in as list, which only happens in our case for an in-line value and thus skip the url or path check

ajbozarth:
@maximilien @akihikokuroda @george-lhj I believe I have addressed all the review and comments made so far, is there any further discussion or review we want to handle before merging this and cutting a release?

ajbozarth:
> ```# this assumes `scripts/code-snippet.py is relative to where YAML file located alternatively URI could be file://...```

how should we determine what relative to the yaml is? currently we pass in the yaml file and each command parses it using a utility function. Should we update each place it's passed in to save the yaml dir for later reference at run? or should we update the yaml parsing utility to save the value somewhere in the agent object? If the later where in the yaml should we save it?

ajbozarth:
As for your suggestion in https://github.com/AI4quantum/maestro/pull/671#issuecomment-3141151782:
- the first option works with this implementation (except what it's relative to, noting my comment above)
- the third option kind of works with this, but you have to have a raw url so it finds the code rather than a web page. If we want to support just putting a GitHub link it would require adding handling to extrapolate the raw url to then fetch.
- which brings me to option two, this would need the same extrapolation mentioned above, but I'm not sure I like this option, it seems unnecessarily verbose when a user could instead just give a url.

Per your followup comment I'll start work on tracking the agent yaml dir path and coordinate with @akihikokuroda

ajbozarth:
> support GitHub file urls directly, not require users to provide the raw url

just pushed a solution for this, will work on the other task tomorrow

ajbozarth:
> make the paths in the yaml relative to the yaml file

pushed an update for this as well, this finishes addressing the first and third cases in https://github.com/AI4quantum/maestro/pull/671#issuecomment-3141151782

Unless desired by others I don't plan to address case 2

ajbozarth:
> can you update the PR description (if necessary) to capture any behavior / functionality changes that have arisen in the course of reviews? 

done

docs/USER_GUIDE.md
```diff
@@ -593,6 +594,43 @@ spec:
 🐝 Response from countdown: happy
 ```
 
+### Simple code agent
```
psschwei, line 597: nit: the doc update seems unrelated to the PR, so in the future I would just do a separate PR to add the doc changes (will keep PRs smaller / easier to review)

ajbozarth: This doc update was requested by @george-lhj he wanted me to document how `spec.code` works, but we didn't even have a section on code agents to add it to, so I added the section and included the documentation for `spec.code` in it just like I did for llm agents earlier in the same doc

maximilien (commented):
LGTM but I know others had comments so letting them give you final LGTM and merge

src/maestro/agents/utils.py
ajbozarth: passing in a list of string seems to only happen in tests when constructing the agent scheme manually instead of using the import, still leaving this here for compatibility

ajbozarth:
> Would you add source_file to [schema](https://github.com/AI4quantum/maestro/blob/main/src/maestro/schemas/agent_schema.json)

I added it to the schema, but are we sure we want to? That value is only added to the object after it is read in and is not part of the yaml file itself. @akihikokuroda

akihikokuroda (approved):
LGTM!

ajbozarth:
@maximilien can't merge since you have an outstanding "requested changes" review

PR #675, opened 2025-07-30
bug: update to only one trace

### LINKED ISSUE #674: bug: scoring agent introduces 3 traces instead of 1
Currently each workflow run when using scoring logs 3 separate traces (an overall, hallucination, and a relevancy) trace. Need it convert it to just one trace, with these metrics as metadata so its easy to format in the future.

george-lhj:
fix #674

natively, Opik doesn't support creating just one trace because they want to give you context on why the scores were given. 
created a workaround so that there is only one trace, and has all the metrics and reasons as metadata. 

switched to qwen for evaluation

### DIFF
--- a/tests/agents/scoring_agent/agents.yaml
+++ b/tests/agents/scoring_agent/agents.yaml
@@ -5,11 +5,11 @@ metadata:
   labels:
     app: testapp
 spec:
-  model: deepseek-r1:latest
+  model: qwen3:latest
   description: Simple test agent
   instructions: Answer the question truthfully
-  framework: beeai
-  mode: remote
+  framework: openai
+  mode: local
 ---
 apiVersion: maestro/v1alpha1
 kind: Agent
--- a/tests/agents/scoring_agent/workflow.yaml
+++ b/tests/agents/scoring_agent/workflow.yaml
@@ -12,7 +12,7 @@ spec:
     agents:
       - simple_test
       - score
-    prompt: Is the most well known Paris the one in the United States?
+    prompt: Is the most well known "Paris" the city in the United States?
     steps:
       - name: answer
         agent: simple_test
--- a/tests/yamls/agents/scoring_agent.yaml
+++ b/tests/yamls/agents/scoring_agent.yaml
@@ -6,8 +6,8 @@ metadata:
     app: test-example
 spec:
   model: qwen3:latest
-  framework: beeai
-  mode: remote
+  framework: openai
+  mode: local
   description: this is a test
   instructions: Concisely answer the question to the best of your ability, and if you are unsure, say "I don't know". 
 
@@ -20,8 +20,8 @@ metadata:
     app: test-example
 spec:
   model: qwen3:latest
-  framework: beeai
-  mode: remote
+  framework: openai
+  mode: local
   description: this is a test
   instructions: You are a gibberish generator. You will be given a prompt and you will generate a gibberish response that is not related to the prompt at all.
 


george-lhj:
Im not sure why the tests are failing, from the logs it seems all of them pass

ajbozarth:
The test failure is relevant, it just the CI is not printing the actual error in the log, when run locally I get: 
```bash
____________________________________________________________________________________________________ test_metrics_agent_run_with_context _____________________________________________________________________________________________________

monkeypatch = <_pytest.monkeypatch.MonkeyPatch object at 0x1285e1d90>

    def test_metrics_agent_run_with_context(monkeypatch):
        seen = {"relevance": None, "hallucination": None}
    
        class DummyScore:
            def __init__(self, value):
                self.value = value
    
        def fake_rel(self, input, output, context):
            seen["relevance"] = context
            return DummyScore(0.50)
    
        def fake_hall(self, input, output, context):
            seen["hallucination"] = context
            return DummyScore(0.20)
    
        monkeypatch.setattr(AnswerRelevance, "score", fake_rel)
        monkeypatch.setattr(Hallucination, "score", fake_hall)
    
        printed = []
        monkeypatch.setattr(ScoringAgent, "print", lambda self, msg: printed.append(msg))
    
        agent_def = {
            "metadata": {"name": "metrics_agent", "labels": {}},
            "spec": {
                "framework": "custom",
                "model": "qwen3:latest",
                "description": "desc",
                "instructions": "instr",
            },
        }
        agent = ScoringAgent(agent_def)
        prompt = "What is the capital of France?"
        response = "Lyon"
        context = ["Paris is the capital of France."]
>       out = asyncio.run(agent.run(prompt, response, context=context))
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

tests/agents/test_scoring_agent.py:55: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../../../../.local/share/uv/python/cpython-3.12.8-macos-aarch64-none/lib/python3.12/asyncio/runners.py:194: in run
    return runner.run(main)
           ^^^^^^^^^^^^^^^^
../../../../.local/share/uv/python/cpython-3.12.8-macos-aarch64-none/lib/python3.12/asyncio/runners.py:118: in run
    return self._loop.run_until_complete(task)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
../../../../.local/share/uv/python/cpython-3.12.8-macos-aarch64-none/lib/python3.12/asyncio/base_events.py:686: in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <maestro.agents.scoring_agent.ScoringAgent object at 0x1285e2180>, prompt = 'What is the capital of France?', response = 'Lyon', context = ['Paris is the capital of France.']

    async def run(
        self, prompt: str, response: str, context: list[str] | None = None
    ) -> any:
        """
        Args:
          prompt:   the original prompt
          response: the agent's output
          context:  optional list of strings to use as gold/context
    
        Note: The response only supports strings for now because Opik's evaluation passes in this as a json object.
        Currently anything else is unsupported, so we can avoid python crash but the Opik backend itself will fail.
    
        Returns:
          The original response (unchanged).  Metrics are printed to stdout.
        """
        assert isinstance(response, str), (
            f"ScoringAgent only supports string responses, got {type(response).__name__}"
        )
        response_text = response
        ctx = context or [prompt]
    
        # Temporarily disable Opik tracing so metric calls don't create traces
        original_disable = os.environ.get("OPIK_TRACK_DISABLE", "false")
        os.environ["OPIK_TRACK_DISABLE"] = "true"
    
        try:
            answer_relevance = AnswerRelevance(model="ollama/qwen3")
            hallucination = Hallucination(model="ollama/qwen3")
    
            rel_eval = answer_relevance.score(prompt, response_text, context=ctx)
            hall_eval = hallucination.score(prompt, response_text, context=ctx)
    
            rel_value = rel_eval.value
            hall_value = hall_eval.value
>           rel_reason_raw = rel_eval.reason
                             ^^^^^^^^^^^^^^^
E           AttributeError: 'DummyScore' object has no attribute 'reason'

src/maestro/agents/scoring_agent.py:64: AttributeError

```

src/maestro/agents/scoring_agent.py
```diff
@@ -48,31 +48,71 @@ async def run(
...
+                ", ".join(rel_reason_raw)
+                if isinstance(rel_reason_raw, (list, tuple))
+                else rel_reason_raw or ""
+            )
+            hall_reason = (
+                ", ".join(hall_reason_raw)
+                if isinstance(hall_reason_raw, (list, tuple))
+                else hall_reason_raw or ""
+            )
+        finally:
```
psschwei, line 76: I think we dropped the `except` block for this... is that ok?

src/maestro/agents/scoring_agent.py
```diff
@@ -48,31 +48,71 @@ async def run(
...
 
+        # Temporarily disable Opik tracing so metric calls don't create traces
+        original_disable = os.environ.get("OPIK_TRACK_DISABLE", "false")
+        os.environ["OPIK_TRACK_DISABLE"] = "true"
+
         try:
-            answer_relevance = AnswerRelevance(model=self._litellm_model)
-            hallucination = Hallucination(model=self._litellm_model)
+            answer_relevance = AnswerRelevance(model="ollama/qwen3")
+            hallucination = Hallucination(model="ollama/qwen3")
```
psschwei, line 57: Why hard code here? Is it not possible to define the model when creating the agent?

src/maestro/workflow.py
```diff
@@ -201,8 +242,37 @@ async def _condition(self):
...
+
         return {"final_prompt": prompt, **step_results}
 
+    def _create_workflow_trace(self, initial_prompt, final_prompt, step_results):
+        """
+        Create a single trace for the entire workflow run with scoring metrics as metadata.
+        Only creates traces if Opik is initialized (i.e., when there's a scoring agent).
+        """
+        if self._opik is None:
+            return
```
psschwei, line 255: out of curiosity, are there cases where this could be called if opik isn't initialized?

src/maestro/agents/scoring_agent.py
george-lhj: fixed

src/maestro/agents/scoring_agent.py
george-lhj: fixed

src/maestro/workflow.py
george-lhj: I think this check is necessary because the process is as follows:
1) _create_workflow_trace gets called every time a workflow runs (both success and error cases)
2) self._opik is only set to a real Opik() object when there's a scoring agent. When there's no scoring agent, self._opik stays as None
3) Without the check, the code tries to call None.trace() which crashes

george-lhj: I was running _create_workflow_trace every time because I wasn’t sure whether we ultimately want to keep scoring and logging decoupled.

My assumption was that it might be better design-wise to decouple them, for example, we might want to log traces for all workflows even if we're not scoring them. Calling _create_workflow_trace unconditionally (I think) makes it easier to support that kind of behavior down the line without needing to restructure control flow later.

Still a bit unclear in my head tbh, so welcome to your suggestions

src/maestro/agents/scoring_agent.py
```diff
@@ -48,31 +48,74 @@ async def run(
...
+                else hall_reason_raw or ""
+            )
+        except Exception as e:
+            self.print(f"[ScoringAgent] Warning: could not calculate metrics: {e}")
+            return {"prompt": response_text, "scoring_metrics": None}
+        finally:
+            if original_disable == "false":
+                os.environ.pop("OPIK_TRACK_DISABLE", None)
+            else:
+                os.environ["OPIK_TRACK_DISABLE"] = original_disable
```
maximilien, line 83: Explain this logic a bit. Seems odd to be popping env vara in finally

src/maestro/agents/scoring_agent.py
```diff
@@ -25,7 +26,6 @@ def __init__(self, agent: dict) -> None:
             self._litellm_model = raw_model
         else:
             self._litellm_model = f"ollama/{raw_model}"
-        self._opik = Opik()
 
     async def run(
```
maximilien, line 30: This method should be split into smaller sections (private) to make it readable and testable

src/maestro/workflow.py
```diff
@@ -129,6 +136,37 @@ def _create_or_restore_agents(self):
 
                 self.agents[agent_name] = agent_instance
 
+        # Check if there's a scoring agent in the workflow and initialize Opik only if needed
```
maximilien, line 139: Same here in terms of breaking into smaller private methods that can be purposeful and named appropriately

maximilien (changes requested):
Code could be simplified. I'd be curious if you ran lint (ruff) and what it would say. Could also use AI to ask to simplify... also tests :)

src/maestro/agents/scoring_agent.py
george-lhj: ya, I originally want to make sure that the disable is true when it starts, but this logic here isn't necessary so i removed it

george-lhj:
@ajbozarth if you have time to review this, i switched this to openai workflow so its easy to test and run. The only thing you probably need is an Opik Key to access the dashboard. If you don't want to create an account I think you should be able to trace the logs to confirm if its working correctly also

PR #676, opened 2025-07-30
test: improve test output in CI

ajbozarth:
While tracking down an error in #675 we found that the pytest failures summary isn't shown in the CI, this updates the CI to show the results summary.

In addition, while forging out how to display the summary I noticed that three test that leverage the `deployments/Dockerfile` were silently failing. I updated the code so the failures are no longer silent, but then added a pytest skip (as was used by the other deploy tests). I will open a separate follow up issue to get the deploy tests passing (as fixing the Dockerfile was non-trivial)

### DIFF
--- a/.github/workflows/maestro_run-tests.yaml
+++ b/.github/workflows/maestro_run-tests.yaml
@@ -33,4 +33,4 @@ jobs:
     - name: Run tests
       run: |
         export PYTHONPATH=$PYTHONPATH:$(pwd)/src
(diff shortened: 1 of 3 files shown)


ajbozarth:
I've opened https://github.com/AI4quantum/maestro/issues/677 to follow up on the test failures I set to skip

george-lhj (approved):
lgtm for the non docker tests

PR #681, opened 2025-08-01
feature: enable workflow streaming in api

### LINKED ISSUE #682: feature: streaming endpoint in maestro
Why We Need a New Endpoint in Maestro

Currrently building a frontend that streams agent responses to the user in real time. Right now, we run the builder backend using maestro serve, and the frontend sends user prompts to our own /api/chat_builder_agent endpoint, which internally makes a POST request to http://localhost:8000/chat.

This /chat endpoint connects to a single LLM agent (e.g., BeeAI) and returns a streamed response. However, it bypasses the actual Workflow.run() logic, which coordinates multiple agents in sequence with dependencies and logic defined in a YAML workflow.

So, the current flow only streams raw output from a single LLM, not the full orchestrated result from the Maestro workflow. Even though Workflow.run() supports streaming internally (when called via maestro run workflow.yaml), there is no exposed streaming API for that path.
This means our users don’t see the value of orchestration only the final response, without the full reasoning, logic, or intermediate agent outputs.

We should expose a new endpoint, for example, /chat_builder_stream,  that:

Internally runs Workflow.run(prompt) with streaming enabled.
Streams back the full multi-agent output, similar to what happens when running maestro run workflow.yaml so that as new output gets returns from LLM, it posts to the api.

CC: @maximilien

george-lhj:
fixes #682 enables the streaming of answers inside of maestro serve, this allows us to display the process inside of builder

### DIFF
--- a/tests/integration/serve/test_streaming.py
+++ b/tests/integration/serve/test_streaming.py
@@ -18,34 +18,40 @@ import json
 import time
 import requests
 import subprocess
-import sys
+import pytest
 
 
-class StreamingTester:
+class TestStreaming:
     """Self-contained streaming tester."""
 
-    def __init__(self):
-        self.base_url = "http://localhost:8000"
-        self.server_process = None
-        self.test_results = []
+    @classmethod
+    def setup_class(cls):
+        """Set up test class."""
+        cls.base_url = "http://127.0.0.1:8000"
+        cls.server_process = None
+        cls.test_results = []
 
-    def start_server(self):
+        if not cls.start_server():
+            pytest.fail("Failed to start server")
+
+    @classmethod
+    def start_server(cls):
         """Start the Maestro workflow server."""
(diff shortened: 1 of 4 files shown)


ajbozarth:
I'd like to test this with our internal use case, but I won't have time till Monday. Code-wise, on a short read through I don't see any glaring issues

tests/integration/serve/test_streaming.py
```diff
@@ -0,0 +1,251 @@
...
+"""
+
+import json
+import time
+import requests
+import subprocess
+import sys
+
+
+class StreamingTester:
```
akihikokuroda, line 24: Would you change the class name to start with `Test` so that pytest picks this up?

akihikokuroda (approved):
LGTM.  Thanks!

PR #686, opened 2025-08-05
feat: Enable maestro-knowledge tool for maestro agent

akihikokuroda:
Enable maestro-knowledge for maestro agents.

The maestro-knowledge (ToolHive) MCP server is deployed in the kubernetes cluster by

```cmd
maestro create tests/yamls/tools/maestro_knowlege_mcp.yaml
```

The maestro agent can access to the tool by adding `maestrok` to the `tools` list in the agent yaml file.

### DIFF
--- a/tests/yamls/tools/maestro_knowlege_mcp.yaml
+++ b/tests/yamls/tools/maestro_knowlege_mcp.yaml
@@ -1,0 +1,34 @@
+apiVersion: maestro/v1alpha1
+kind: MCPTool
+metadata:
+  name: maestrok
+  namespace: default
+spec:
+  image: ghcr.io/akihikokuroda/maestro-knowledge:latest
+  transport: streamable-http
+  targetPort: 8030
+  permissionProfile:
+    type: builtin
+    name: network
+  resources:
+    limits:
+      cpu: '100m'
+      memory: '128Mi'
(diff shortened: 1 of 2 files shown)


ajbozarth:
Thanks for this example, I'll take a deep look at this in conjunction with our internal use case tomorrow

tests/yamls/tools/maestro_knowlege_mcp.yaml
```diff
@@ -0,0 +1,34 @@
+apiVersion: maestro/v1alpha1
+kind: MCPTool
+metadata:
+  name: maestrok
```
ajbozarth, line 4: nit: I believe we've been using `maestro-k` as the short name for `maestro-knowledge`

ajbozarth:
IIUC using this tool def an agent should be able to query the maestro-knowledge mcptool given the correct agent instructions to do so?

akihikokuroda:
@ajbozarth Yes, it is.  Maestro creates the maestro knowledge mcp server in a kubernetes cluster.  It creates one deployment and one service picking up the container image from registry.

ajbozarth (approved):
Per my last comment we should open a follow ups with actual tests if we merge this without them

PR #689, opened 2025-08-06
bug: track all models used in scoring

### LINKED ISSUE #688: bug: scoring log metadata
The metadata needs to capture the models used. Currently it only captures the scoring agent + its model used. However, it should capture the model of all the agents used in the entire workflow.

george-lhj:
Fixed #688 dashboard model tracking - Now properly separates workflow models from scoring models instead of incorrectly showing scoring agent's model as the workflow model

Added workflow model metadata - Tracks which models each workflow agent uses (e.g., simple_test: gpt-oss:latest) in a clean key-value format

Cleaned up scoring metrics - Removed redundant faithfulness metric and scoring_agent field, renamed provider to framework_provider for better semantics

### DIFF
--- a/src/maestro/agents/agent.py
+++ b/src/maestro/agents/agent.py
@@ -136,6 +136,8 @@ def restore_agent(agent_name: str):
     Restore agent from storage.
     """
     agents = _load_agent_db()
+    if agent_name not in agents:
+        return agent_name, False
     agent_data = agents[agent_name]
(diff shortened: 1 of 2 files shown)


george-lhj:
image
New data output structure

ajbozarth (approved):
code LGTM, but I haven't tested locally

PR #690, opened 2025-08-07
feature: add more eval metrics + docs

george-lhj:
I've added more relevant metrics that a user might want to see when trying to improve a prompt, including response time/workflow time, and also tokens. 

Additionally, I have added documentation on how to use/add the scoring agent into any workflow.

@psschwei this might be helpful for you to look at and test out, since you'll be able to take a look at the overall metadata scheme


george-lhj:
I don't think my PR is related to the kind test failure, because its related to adding metrics (so if anything it should fail my run tests).

src/maestro/agents/openai_agent.py
```diff
@@ -242,6 +332,64 @@ def _process_agent_result(self, result: Optional[Any]) -> str:
...
+                    self.response_tokens = usage.completion_tokens
+                if hasattr(usage, "total_tokens"):
+                    self.total_tokens = usage.total_tokens
+                self.print(
+                    f"INFO [OpenAIAgent {self.agent_name}]: Extracted token usage from result - Prompt: {self.prompt_tokens}, Response: {self.response_tokens}, Total: {self.total_tokens}"
+                )
+                return
+
+            if hasattr(result, "messages"):
+                for message in result.messages:
```
maximilien, line 352: I'd move these massive if checks into their own private method. And if these are similarly used other places then put in a utility class or file in utils package that could be reused. Basically search code base for duplicates or similar code

src/maestro/workflow.py
```diff
@@ -468,6 +490,75 @@ def _initialize_opik(self) -> None:
...
+            "total_tokens": 0,
+            "agent_token_usage": {},
+        }
+
+        for agent_name, agent in self.agents.items():
+            if hasattr(agent, "get_token_usage"):
+                token_usage = agent.get_token_usage()
+                total_token_usage["agent_token_usage"][agent_name] = token_usage
+
+                if "prompt_tokens" in token_usage:
```
maximilien, line 549: This seems to be duplicated also. Move to a common method and use. Avoid repeating yourself :)

maximilien (changes requested):

george-lhj:
seems like my local demo tests are running correctly, i dont know how this is affeccting operator, as it passed in bc357fa (and i havent made any changes since essentially)

akihikokuroda:
The operator failure is caused by my bad implementation :-(. Please add   "tiktoken" around line 70 in operator/Dockerfile-engine.

akihikokuroda:
This is the error.  I see this sometimes.  It's confusing the module name.  Both `openai` and `maestro` have `agent` module.  
 That's causing this error.  I don't know how to fix this unless changing the maestro `agent` to something else.

```
ImportError: cannot import name 'Function' from 'openai.types.chat.chat_completion_message_tool_call' (/usr/local/lib/python3.12/site-packages/openai/types/chat/chat_completion_message_tool_call.py)
```

PR #699, opened 2025-08-13
bug: recreate openai and dspy agents on restore in serve

### LINKED ISSUE #697: bug: `maestro serve AGENT_FILE` fails for openai and dspy agents
While testing #694 I ran into the following bug:

In [`save_agent`](https://github.com/AI4quantum/maestro/blob/cc84a94648099c3b8463be032b3314ea6d1aa871/src/maestro/agents/agent.py#L186) if `pickle.dumps` fails it will save the agent in the DB as a dict instead of a class.
This seems to work fine in most use cases, but when running `maestro serve AGENT_FILE` it will fail with `Error: Error in chat endpoint: 'dict' object has no attribute 'run'` on https://github.com/AI4quantum/maestro/blob/696f4a4e227defa620c1d5c91113f984240ad513/src/maestro/cli/fastapi_serve.py#L104 when it expects a class, but receives a dict.

We should either fix the cause of the pick failures or update serve to created a agent class from the dictionary before calling run.

For reference these are the pickle errors:
- for openai: 
    ```
    cannot pickle '_thread.RLock' object
    ```
- for dspy:
    ```
    Can't pickle StringSignature(user_request -> process_result
        instructions='<instructions str>'
        user_request = Field(annotation=str required=True json_schema_extra={'__dspy_field_type': 'input', 'prefix': 'User Request:', 'desc': '${user_request}'})
        process_result = Field(annotation=str required=True json_schema_extra={'desc': 'Message that summarizes the process result, and the final answer to the user questions and requests.', '__dspy_field_type': 'output', 'prefix': 'Process Result:'})
    ): attribute lookup StringSignature on dspy.signatures.signature failed
    ```

akihikokuroda:
fix #697

### DIFF
--- a/src/maestro/cli/fastapi_serve.py
+++ b/src/maestro/cli/fastapi_serve.py
@@ -14,7 +14,7 @@ from fastapi.middleware.cors import CORSMiddleware
 from fastapi.responses import StreamingResponse
 from pydantic import BaseModel
 
-from maestro.workflow import create_agents, Workflow
+from maestro.workflow import create_agents, Workflow, get_agent_class
 from maestro.agents.agent import restore_agent
 from maestro.cli.common import parse_yaml, Console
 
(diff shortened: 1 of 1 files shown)


ajbozarth (approved):
I have confirmed that this fixes the error when serving a openai or dspy agent

Issue #474, opened 2025-05-08
question: why slack_agent.py and not a slack_agent.yaml


maximilien:
I think `src/agents/slack_agent.py` should be defined as a YAML agent and not in code.

Please elaborate why if you disagree. I can elaborate my reasoning if needed but if we agree then let's use this issue to make that change.

akihikokuroda:
Maestro didn't have an agent support that easily use mcp tools a few weeks ago.  Now Openai agent supports mcp tools that includes posting messages to slack.  So it can be done.

akihikokuroda:
This hasn't done yet.  To make this an agent, the key to the slack must be given to the LLM.  I'm not sure if it's a good idea.  @maximilien WDYT?

akihikokuroda:
Slack MCP tool can replace this.  I'll deprecate Slack agent.

PR #703, opened 2025-08-20
feat: Adding Slack message posting MCP server

akihikokuroda:
#474 

This PR has slack message posting MCP server and one workflow example that show the usage of the mcp server.  The slack custom agent can be replaced with this.


mcp/examples/slack/agents.yaml
```diff
@@ -0,0 +1,14 @@
...
+kind: Agent
+metadata:
+  name: slack
+  labels:
+    app: slack
+spec:
+  model: llama3.1:latest
+  description: post slack message
+  instructions: You are a helpful agent. You have a slack tool.
+  framework: beeai
```
george-lhj, line 11: i think we probably should switch to openai especially in the examples, bc most people on the team don'r have beeai setup (and takes a long time)

PR #706, opened 2025-08-21
chore: add crewai agent back

### LINKED ISSUE #650: bug: beeai-agent and crewai dependency conflict

akihikokuroda:
fix #650

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -32,6 +32,7 @@ dependencies = [
     "dspy>=2.6.27",
     "kubernetes>=33.1.0",
     "ddgs>=9.4.0",
+    "crewai>=0.134.0",
 ]
 
 [dependency-groups]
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -32,6 +32,7 @@ dependencies = [
     "dspy>=2.6.27",
     "kubernetes>=33.1.0",
     "ddgs>=9.4.0",
+    "crewai>=0.134.0",
```
ajbozarth, line 35: do we no longer need the dependency group like before (https://github.com/AI4quantum/maestro/pull/651/files#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711L36)? If so we should update the docs and remove the extra handling in a followup

ajbozarth (approved):
Pending an answer to my comment this LGTM

pyproject.toml
akihikokuroda: We don't need it at least for now.  Updating docs in a follow up.

PR #725, opened 2025-09-22
chore: test python3.13

### LINKED ISSUE #655: chore: include install instructions for system Python v3.13
Python 3.13 is the system version on some systems, which throws an error when trying to install Maestro (as 3.13 is not supported). The install instructions should be updated to include guidance for what to do in this situation.

george-lhj:
fixes #655


ajbozarth:
> tested locally and 3.13 is actually compatiable w/maestro

Did you test with beeai? IIRC beeai was the reason we locked to 3.12 originally

PR #763, opened 2025-10-14
docs: add pip install uv step to virtual environment setup

li-krystal-1:
Updates the virtual environment setup instructions in CONTRIBUTING.md to be more beginner friendly. I was trying to set up with the -m venv command but uv is used to run tests later on.


george-lhj:
lgtm

PR #768, opened 2025-10-21
enhancement: use mermaid npm for verification

### LINKED ISSUE #353: bug: verify generated mermaid
Some of the generated mermaid are failing to render. For example, when agent and step names have spaces the mermaid code is not quoted so fails to render. 

Need to find a way to add render to `tools/check-mermaid.sh` so not only is the mermaid generated but also checked.

george-lhj:
fixes #353


maximilien (approved):
LGTM

